PR #201, opened 2016-03-11
MERC-409: Enable theme to hook to build system

hegrec:


mcampa:
:+1: LGTM, I've been using this branch for a few days with no issues. 
We are safe to merge this because is backwards compatible with old themes

PR #204, opened 2016-03-15
MERC-419 - Bundle should not be created on error.

mickr:
* Moved createWriteStream to after the error check to prevent a write stream being created even when there is an error.
* Propagted the error to the callback of `Async.series` to handle the failure and throw to stop execution of tasks.

### DIFF
--- a/bin/stencil-bundle
+++ b/bin/stencil-bundle
@@ -87,7 +87,7 @@ Async.series([
         Async.parallel(tasks, function (err, assembledData) {
             var outputFolder = Program.dest ? Program.dest : themePath,
                 outputName = Program.filename ? Program.filename : defaultName,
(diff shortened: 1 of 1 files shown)


mickr:
@hegrec @mcampa @sherrybc

bin/stencil-bundle
```diff
@@ -102,9 +101,11 @@ Async.series([
                 archive = Archiver('zip');
 
             if (err) {
-                return console.error(err);
+                return cb(err);
             }
 
+            var output = Fs.createWriteStream(Path.join(outputFolder, outputName));
```
hegrec: Should declare output above so the vars are together.

hegrec:
:+1: LGTM after minor comments

mickr:
Updated, will merge when travis is complete.

PR #206, opened 2016-03-22
Allow bundles without lang folders

hegrec:
### What
Allow themes that do not support translations and have no lang folder to be bundled successfully.

EAT-727

@mcampa @mickr

### DIFF
--- a/lib/langAssembler.js
+++ b/lib/langAssembler.js
@@ -1,9 +1,11 @@
 var Async = require('async'),
     Fs = require('fs'),
     Path = require('path'),
-    _ = require('lodash');
+    _ = require('lodash'),
+    localeDirectory = 'lang';
 
 module.exports.assemble = assemble;
+module.exports.localeDirectory = localeDirectory;
 
 /**
  * Assembles together all of the lang files.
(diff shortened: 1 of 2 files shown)


lib/langAssembler.js
```diff
@@ -15,6 +15,14 @@ module.exports.assemble = assemble;
 function assemble(callback) {
     var localeDirectory = 'lang';
 
+    try {
+        Fs.statSync(localeDirectory);
+    }
+    catch (e) {
+        // No lang folder, callback immediately
+        return callback(null, 'no_lang_folder');
```
mcampa: I think you should return an error in the callback and then check the error type in `bin/stencil-bundle:294`

hegrec: It's not really an error though, everything can still proceed, but it's there to provide an informational alert. An error should only be for something that keeps it from bundling.

mcampa: I disagree with that, the return data, should be consistent, if the `results` variable is an object should always be an object. If the callback returns an error shouldn't indicate that the build has failed. You can set an error type or name and check for that name.

You could also check if the file exist before calling `assemble()`

hegrec: Went forward with checking if it exists before calling assemble.

mcampa:
:+1:

PR #207, opened 2016-03-29
MERC-497 and MERC-534 Add manifest.json and Fix for Windows

mickr:
MERC-497 and MERC-534

Added Manifest generation and refactoring bundle process. Tests are coming. @sherrybc @bc-ejoe @hegrec


package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "postinstall": "node make --install",
-    "test": "lab -t 75 -e 'development'",
+    "test": "lab -t 75 -l -e 'development'",
```
mickr: JSPM leaks global variables like a sieve. Disabling leak detection for now.

bin/stencil-bundle
```diff
@@ -72,261 +62,14 @@ if (stencilConfig && typeof(stencilConfig.production) === 'function') {
...
             callback();
         });
-
     }
 }
 
-
 // TODO: Remove explicit support here in the platform tool and move this to the theme.
 if (configuration.jspm) {
+    console.log(configuration.jspm);
```
hegrec: Remove console log?

mickr: Woops

bin/stencil-bundle
```diff
@@ -57,13 +43,17 @@ if (!themeConfig.configExists()) {
 
 configuration = themeConfig.getRawConfig();
 
+var bundle = new Bundle(themePath, themeConfig, configuration);
+
+
 if (configuration.css_compiler) {
-    tasks.css = getCssAssembleTask(configuration.css_compiler);
+    tasks.css = bundle.getCssAssembleTask(configuration.css_compiler);
```
hegrec: Any way this can be put into the new class as an implementation detail?

bin/stencil-bundle
```diff
@@ -57,13 +43,17 @@ if (!themeConfig.configExists()) {
...
 }
 
-tasks.templates = assembleTemplatesTask;
-tasks.lang = assembleLangTask;
-tasks.schema = assembleSchema;
+tasks.templates = bundle.assembleTemplatesTask.bind(bundle);
```
hegrec: Same here, could this all be put into the bundle class with a single exposed `bundleTheme`?

lib/stencilBundle.js
```diff
@@ -0,0 +1,321 @@
...
+var Path = require('path');
+var BundleValidator = require('../lib/bundleValidator');
+var Cycles = require('../lib/Cycles');
+var CssAssembler = require('../lib/cssAssembler');
+var LangAssembler = require('../lib/langAssembler');
+var TemplateAssembler = require('../lib/templateAssembler');
+var validator;
+
+function Bundle(themePath, themeConfig, rawConfig, options) {
+    validator = new BundleValidator(themePath, themeConfig, Program.marketplace !== true);
```
hegrec: The Program parameters should be passed in from the bin file to keep the concern of being a command line program out of the class.

bin/stencil-bundle
mickr: That should be pretty easy to do

lib/stencilBundle.js
mickr: I did that at first then found command runs as a singleton, but the feedback makes sense and reassures I had the correct approach before. I'll switch it back tomorrow

bin/stencil-bundle
```diff
@@ -57,277 +42,7 @@ if (!themeConfig.configExists()) {
...
-            minify: true,
-            mangle: true
-        }).then(function () {
-            console.log('ok'.green + ' -- JavaScript Bundling Finished');
-            console.error = oldConsoleError;
-            callback(null, true);
-        });
-    }
-}
+var bundle = new Bundle(themePath, themeConfig, configuration, stencilConfig, {cmd: Program});
```
hegrec: should declare variable up above and maybe put each argument on its own line because there are so many

mickr:
@bc-ranji this needs to be tested on Windows and Mac.

mickr:
I uploaded a private bundle from Windows to verify the Windows fix.

hegrec:
:+1:

mickr:
Rebased, just waiting on Travis

mcampa:
I see this `manifest.json` file now hanging in the stencil directory? should this be added to `.gitignore`?
Why a file is created? should the file only be inside the bundle? if that is the case, you don't need to create the file at all. I mean you can zip the file without creating the file

@mickr

lib/stencilBundle.js
```diff
@@ -0,0 +1,357 @@
...
+        if (err) {
+            return callback(err);
+        }
+
+        filePaths = filePaths.map(function (file) {
+            return upath.toUnix( file.replace(self.templatesBasePath + Path.sep, '').replace('.html', ''));
+        });
+
+        output.templates = filePaths;
+        Fs.writeFile(self.themePath + '/manifest.json', JSON.stringify(output), function (err) {
```
mcampa: I think you should have use `archive.append()` instead of creating a file

mickr: I can put in a quick PR for that.

PR #217, opened 2016-04-19
Allow stencil-start to run on case-sensitive file systems.

jnorth:
On case-sensitive file systems, the `stencil-start` script would cause an error when importing the Cycles module.


bin/stencil-start
```diff
@@ -2,7 +2,7 @@
 
 var _ = require('lodash');
 var Bs = require('browser-sync').create();
-var Cycles = require('../lib/cycles');
+var Cycles = require('../lib/Cycles');
```
mickr: Can you rename the file to `cycles.js` and test it, that will put it more within standards.

jnorth: Good call 👍

mickr:
👍

hegrec:
👍

PR #221, opened 2016-05-05
MERC-681 Improved Theme Editor build process by removing JSPM and using Webpack

mcampa:
- Removed jspm for theme editor
- Removed the `make` file that zip all the jspm dependencies
- Added webpack to bundle `theme-editor` and all dependencies
- The sdk is also bundled into one file including `jschannel` and `js-cookie`
- Much more simple `stencil-editor` plugin (no scanning `package.json` for paths)

One extra benefit is that, you don't need to "transclude directories" anymore when working with Theme Editor/CLI. Just use `npm link`

@deini @hegrec @mickr @bc-jstoffan @sherrybc @bc-ejoe


mickr:
Looks good, but will wait for @deini or @hegrec to take a look too

package.json
```diff
@@ -68,41 +67,30 @@
     "wreck": "^5.4.0"
   },
   "devDependencies": {
+    "angular": "^1.5.5",
```
deini: We use angular `1.3.7` so all the angular dependencies need to be changed to that. Also we should lock everything too.

package.json
```diff
@@ -68,41 +67,30 @@
...
-      "babel-runtime": "npm:babel-runtime@^5.1.13",
-      "core-js": "npm:core-js@^0.9.4"
-    },
-    "overrides": {
-      "github:bigcommerce-labs/bcapp-pattern-lab@1.17.6": {
-        "format": "global"
-      }
-    }
+    "ng-common": "bigcommerce-labs/ng-common",
+    "ng-stencil-editor": "bigcommerce-labs/ng-stencil-editor",
```
deini: Does this work? Don't you have to give it the github ssh url and also point it to a specific version?

mcampa: I guess the new npm supports this format now. I just ran `npm i --save-dev bigcommerce-labs/ng-stencil-editor`

Let me check, and I'll add the whole url if it does not work for older versions

hegrec:
👍

mcampa:
@bc-ranji This branch is ready for testing, please checkout `mcampa:MERC-681-webpack` to test. Let me know how it goes

bc-ranji:
💚

PR #224, opened 2016-05-26
MERC-758 - Revert MERC-683

mickr:
MERC-758 - Revert MERC-683


mickr:
@mcampa @hegrec pushing this out to fix development. Will have a new PR coming through for bundling only.

mcampa:
👍

PR #225, opened 2016-06-01
Redirects not working when transfer-encoding is present in the headers

mcampa:
@mickr @hegrec

### DIFF
--- a/server/plugins/renderer/responses/redirect-response.js
+++ b/server/plugins/renderer/responses/redirect-response.js
@@ -8,7 +8,9 @@ module.exports = function (location, headers, statusCode) {
         response.statusCode = statusCode;
 
         _.each(headers, function (value, name) {
-            response.header(name, value);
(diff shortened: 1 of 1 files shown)


server/plugins/renderer/responses/redirect-response.js
```diff
@@ -8,7 +8,9 @@ module.exports = function (location, headers, statusCode) {
         response.statusCode = statusCode;
 
         _.each(headers, function (value, name) {
-            response.header(name, value);
+            if (name !== 'transfer-encoding') {
```
mickr: Was this introduced with a newer version of node?

mickr:
👍

server/plugins/renderer/responses/redirect-response.js
mcampa: not sure. all redirects were not responding in my machine. Only until I remove this header started working.

PR #226, opened 2016-06-01
MERC-758

mickr:
* Add option to css assembler for bundling. When creating a bundle all CSS files will normalized to unix paths.

@mcampa @sherrybc

### DIFF
--- a/lib/css-assembler.js
+++ b/lib/css-assembler.js
@@ -2,7 +2,7 @@ var _ = require('lodash'),
     Async = require('async'),
     Fs = require('fs'),
     Path = require('path'),
-    upath = require('upath'),
+    Upath = require('upath'),
     importRegex = /@import\s+?["'](.+?)["'];/g,
     internals = {};
 
@@ -90,7 +90,7 @@ function assemble(cssFiles, absolutePath, type, options, callback) {
 
                     //Ensure all import paths are Unix paths for prod compat.
                     if (options.bundle) {
-                        cssFile = upath.toUnix(cssFileAlias + ext);
+                        cssFile = Upath.toUnix(cssFileAlias + ext);
                     } else {
                         cssFile = cssFileAlias + ext;
                     }
--- a/lib/css-assembler.js
+++ b/lib/css-assembler.js
@@ -2,6 +2,7 @@ var _ = require('lodash'),
     Async = require('async'),
     Fs = require('fs'),
     Path = require('path'),
+    upath = require('upath'),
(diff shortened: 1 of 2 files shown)


lib/css-assembler.js
```diff
@@ -2,6 +2,7 @@ var _ = require('lodash'),
     Async = require('async'),
     Fs = require('fs'),
     Path = require('path'),
+    upath = require('upath'),
```
mcampa: need to add to `package.json`

mcampa: Just to keep consistency capitalize `Upath`

mickr: It is already there - https://github.com/bigcommerce/stencil-cli/blob/master/package.json#L65

lib/stencil-bundle.js
```diff
@@ -82,6 +82,9 @@ Bundle.prototype.initBundle = function () {
 
 Bundle.prototype.getCssAssembleTask = function (compiler) {
     var self = this;
+    var assembleOptions = {
```
mcampa: If this is a constant why do you need to add it as an option and add a 5th parameter?

lib/css-assembler.js
```diff
@@ -76,10 +83,19 @@ function assemble(cssFiles, absolutePath, type, callback) {
...
+                    cssFile;
 
                 if (! err) {
-                    ret[cssFileAlias + ext] = content;
+
+                    //Ensure all import paths are Unix paths for prod compat.
+                    if (options.bundle) {
+                        cssFile = Upath.toUnix(cssFileAlias + ext);
+                    } else {
+                        cssFile = cssFileAlias + ext;
```
mcampa: This code never gets executed because `options.bundle` is always true. I would remove the option parameter and always call Upath.toUnix

lib/css-assembler.js
mcampa: 👍

lib/css-assembler.js
mickr: The options get added in the bundling phase, but this is also called from the `theme-assests` plugin https://github.com/bigcommerce/stencil-cli/blob/master/server/plugins/theme-assets/theme-assets.module.js#L61. When you are running stencil-cli on a Windows machine you don't want the paths converted to Unix because then compilation will fail because the file paths will be wrong and will lead to the issue that this ticket is fixing.

mcampa: I see. I did a quick search for where this method was getting called but I only found one for some reason, my bad.

mcampa:
👍

PR #227, opened 2016-06-02
MERC-769 Custom Templates in Stencil-CLI

mickr:
MERC-769
### Why?
When associating a template to a path the trimRight method was trimming off the last letter of the template name if it ended with h t m l individually and trimRight has since been deprecated in lodash.

* Removing lodash trimRight for just a vanilla replace.
* Change generated format for customLayouts object in stencil-init.

### DIFF
--- a/lib/stencil-init.js
+++ b/lib/stencil-init.js
@@ -14,10 +14,10 @@ internals.parseAnswers = function(JspmAssembler, ThemeConfig, dotStencilFile, do
     // If already set, do nothing otherwise write the empty configurations
     if (!dotStencilFile || dotStencilFile && !dotStencilFile.customLayouts) {
         answers.customLayouts = {
-            'products': {},
-            'search': {},
-            'brands': {},
-            'categories': {}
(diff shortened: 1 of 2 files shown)


mickr:
@mcampa @sherrybc @bc-ranji

mcampa:
👍

lib/stencil-init.js
```diff
@@ -14,10 +14,10 @@ internals.parseAnswers = function(JspmAssembler, ThemeConfig, dotStencilFile, do
...
     if (!dotStencilFile || dotStencilFile && !dotStencilFile.customLayouts) {
         answers.customLayouts = {
-            'products': {},
-            'search': {},
-            'brands': {},
-            'categories': {}
+            'product': {},
+            'page': {},
+            'brand': {},
+            'category': {}
```
sherrybc: 🔤 ?

mickr: Good idea, I'll clean it up

PR #232, opened 2016-06-16
Upgrade stencil editor

hegrec:
@mickr @mcampa @sherrybc @bc-ejoe


server/plugins/stencil-editor/stencil-editor.module.js
```diff
@@ -30,6 +31,7 @@ module.exports.register = function (server, options, next) {
     options.themeServer.ext('onRequest', handlers.onRequest);
 
     // When using stencil-cli variationId = configurationId
+    configurationId = themeConfig.variationIndex + 2;
```
mcampa: why configurationId + 2?

hegrec: why's that?

mcampa: the + 1 thig is only a convention. Because indexes start with 0 but ids start with 1. So an id `1` is index `0`

hegrec: the configId isn't going to be used at all, so I just made it some random thing. I can rename it to be `config`, like how you have `theme` as the versionId

mcampa: https://github.com/bigcommerce/stencil-cli/blob/master/server/plugins/stencil-editor/api/getConfigurations.js#L16

server/plugins/stencil-editor/stencil-editor.module.js
```diff
@@ -53,9 +55,19 @@ module.exports.register = function (server, options, next) {
...
+            handler: function(request, reply) {
+                reply().code(204);
+            }
+        },
+        {
+            method: 'GET',
+            path: '/theme-editor/{versionId}/{variationId}/{configurationId}',
+            config: routesConfig,
+            handler: function(request, reply) {
+                handlers.home(request, reply, themeConfig);
```
mcampa: declare `themeConfig` globally so you don't need to do this

mcampa:
👍

PR #242, opened 2016-09-12
Support definition of custom template mappings using an array

dstaley:
When mapping pages to custom templates locally, developers may want to apply a custom template to more than one URL. This PR modifies the matching algorithm to accept either a string URL, or an array of string URLs to match.

If merged, the following would be valid:
```json
{
  "customLayouts": {
    "product": {
      "featured-product.html": ["/special-product-one", "/special-product-two"],
      "clearance-product.html": "/clearance-product"
    },
    "search": {},
    "brands": {},
    "categories": {}
  }
}
```

Also, I'd love to make sure there's tests for this functionality, but I couldn't quite figure out how to expose the render module's internals to a test script. If anyone could point me in the right direction regarding that, I'd be happy to update the PR with tests!

### DIFF
--- a/server/plugins/renderer/renderer.module.js
+++ b/server/plugins/renderer/renderer.module.js
@@ -346,8 +346,17 @@ internals.getTemplatePath = function (path, data) {
 
     if (internals.validCustomTemplatePageTypes.indexOf(pageType) >= 0 && _.isPlainObject(customLayouts[pageType])) {
         templatePath = _.findKey(customLayouts[pageType], function(p) {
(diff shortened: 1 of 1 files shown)


dstaley:
So, after thinking about this for a while, I'm curious to hear the reason custom templates are defined as `template: url` rather than `url: template`. It seems like the former would be easier to grok, easier to compare against, and also eliminate the need to use arrays (since a url can only have one template, whereas a template can have multiple urls).

mcampa:
👍

PR #243, opened 2016-09-27
MERC-1178 node/hapijs not responding when header 'transfer-encoding' is set

mcampa:
@lord2800 @sherrybc @bc-ong

### DIFF
--- a/server/plugins/renderer/responses/raw-response.js
+++ b/server/plugins/renderer/responses/raw-response.js
@@ -9,17 +9,21 @@ module.exports = function (request, data, headers, statusCode) {
 
     this.respond = function (request, reply) {
         var response;
+        var paylaod = Buffer.isBuffer(data) ? data.toString('utf8') : '';
+
         // TODO This will change when we build the new checkout in SFP.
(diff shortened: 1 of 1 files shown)


lord2800:
Can you point me at some docs about this bug? I'm curious now.

👍 for the code though--some tests would be nice, if possible.

mcampa:
@lord2800 I didn't find any bug in hapijs or node related to this. I'm still not sure why is this happening, I just found that if I don't include that [header](https://en.wikipedia.org/wiki/Chunked_transfer_encoding) it works.

I noticed that in the nodejs [changelog](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V4.md#4.3.0) for 4.3.0 there are some changes to how the headers are parsed

I will try adding some tests

mcampa:
Also, take a look at this: https://github.com/nodejs/node/commit/0eae95eae3#diff-2542201ce4dc4531a266284183f0e166R1811

> Cannot use chunked encoding and a content-length header together
 +           per the HTTP specification

lord2800:
Interesting. Now I'm going to have to do some more digging, because this has all the makings of a bug I'll stumble across later (and be thankful I did my homework on). Thanks!

PR #245, opened 2016-10-21
MERC-1238 jspm fix and paper upgrade

mcampa:
@lord2800 @mattolson


lord2800:
Should we remove node 0.12 from Travis, since we recommend 4+ now? 👍

PR #247, opened 2016-10-21
Require a newer version of node before running

mcampa:
- Require a newer version of node before running
- and bumping the version

@lord2800


lord2800:
👍

PR #248, opened 2016-10-22
Downgrade jspm due to crashing when bundling themes

mcampa:
`stencil bundle` crashes when running `Jspm.bundleSFX()` for no unknown reason.

Downgrading to a known working version while I investigate

@lord2800


mcampa:
This is not a complete revert of the previous change (MERC-1238 is already fixed). It's only a downgrade of jspm

lord2800:
👍

PR #257, opened 2016-11-08
skip problematic files when looking for theme settings while bundling

### LINKED ISSUE #252: `stencil bundle` fails if there is a directory in the project path with `.html` in the name
This can be reproduced by installing stencil-cli to a theme locally, and running `stencil bundle`. The `ThemeConfig.prototype.getSchema` method uses `**/*.html` and doesn't check if it is a file or directory. The `hapi` modules contains the following folder: `node_modules/inert/test/directory/invalid/index.html/`

```
$ stencil bundle
Validating theme...
SCSS Parsing Started...
Template Parsing Started...
Language Files Parsing Started...
Building Theme Schema File...
Generating Manifest Started...
JavaScript Bundling Started...
ok -- Language Files Parsing Finished
ok -- Manifest Generation Finished
ok -- SCSS Parsing Finished
failed  -- EISDIR: illegal operation on a directory, read
/usr/local/lib/node_modules/@bigcommerce/stencil-cli/lib/stencil-bundle.js:78
            throw err;
            ^

Error: EISDIR: illegal operation on a directory, read
    at Error (native)
```

dleavitt:
If run from the theme root, the `stencil bundle` command scans the entirety of the project directory (including everything in node_modules) when looking for HTML files with embedded theme settings.

Is this the desired behavior? In any case, this patch bandaids the problem by skipping (rather than crashing on) files with an html extension that can't be read. It fixes #252, which causes `stencil bundle` to fail, due to a transitive dependency ("inert") containing a directory named "index.html": `node_modules/hapi/node_modules/inert/test/directory/invalid/index.html/`

It also adds some error messages to the stencil bundle command indicating which bundling step failed. It hasn't been tested as the test suite for this module depends on private github repos.

### DIFF
--- a/lib/theme-config.js
+++ b/lib/theme-config.js
@@ -355,12 +355,12 @@ ThemeConfig.prototype.getSchema = function (callback) {
 function fetchThemeSettings(path, callback) {
     var themeSettingsRegexPattern = /\Wtheme_settings\.(.+?)\W/g;
     var themeSettings = {};
-
     Fs.readFile(path, 'utf8', function (err, content) {
         var match;
 
         if (err) {
-            return callback(err);
+            // just skip the file if we can't read it
+            return callback(null, themeSettings);
         }
 
         while (match = themeSettingsRegexPattern.exec(content)) {
--- a/lib/stencil-bundle.js
+++ b/lib/stencil-bundle.js
@@ -98,6 +98,7 @@ Bundle.prototype.getCssAssembleTask = function (compiler) {
             }, function (err, results) {
                 var ret = {};
                 if (err) {
+                    console.log('failed'.red + ' -- %s Parsing');
                     return callback(err);
                 }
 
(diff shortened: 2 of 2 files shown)


lib/theme-config.js
```diff
@@ -355,12 +355,12 @@ ThemeConfig.prototype.getSchema = function (callback) {
...
     var themeSettingsRegexPattern = /\Wtheme_settings\.(.+?)\W/g;
     var themeSettings = {};
-
     Fs.readFile(path, 'utf8', function (err, content) {
         var match;
 
         if (err) {
-            return callback(err);
+            // just skip the file if we can't read it
+            return callback(null, themeSettings);
```
mcampa: Thank you for the contribution, but this is not the right solution. The solution should be to change `'**/*.html'` to `'templates/**/*.html'` on line 321 https://github.com/bigcommerce/stencil-cli/pull/257/files#diff-8da2964f3c728b5ac692995aa59fae52R321

Can you try that and submit the change?

PR #259, opened 2016-11-11
MERC-1493 stencil bundle fails if there is a directory in the project…

mcampa:
… path with .html in the name

Addresses this issue: https://github.com/bigcommerce/stencil-cli/issues/252

@lord2800


lord2800:
👍

PR #260, opened 2016-11-14
MERC-1494 JSPM bug fix

mcampa:
Fixes an error when running stencil init if a file `assets/js/bundle.js` exist

related to https://jira.bigcommerce.com/browse/MERC-1471

@sherrybc @lord2800


lord2800:
👍

PR #264, opened 2017-01-07
MERC-1633 transform bcapp secure cookies to non-secure cookies

mcampa:
## What 
transform bcapp secure cookies to non-secure cookies

## Why
Site-wide HTTPS causes local Theme Editor to get stuck loading when using Stencil-CLI

@lord2800 @mjschock

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bigcommerce/stencil-cli",
-  "version": "1.3.10",
+  "version": "1.3.11",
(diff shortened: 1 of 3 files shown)


mjschock:
LGTM 👍

lord2800:
Hmm. Should we have any concerns for security here? My gut says no since this is `localhost`-only, but I would be remiss if I didn't ask.

mcampa:
It should be fine since this is for localhost

PR #268, opened 2017-01-24
STENCIL-2941 Fix an bug that corrupted images

mcampa:
https://jira.bigcommerce.com/browse/STENCIL-2941

@lord2800 @mjschock


lord2800:
So the fix is to not decode payloads upfront? Just to make sure I'm understanding correctly. If so, 👍 .

mcampa:
We only need to decode payloads that are actually strings, not binary data.

PR #278, opened 2017-02-15
STENCIL-3070 - Expose config version

mjschock:
## WHY
https://jira.bigcommerce.com/browse/STENCIL-3070

### DIFF
--- a/server/plugins/stencil-editor/api/getVariations.js
+++ b/server/plugins/stencil-editor/api/getVariations.js
@@ -19,6 +19,7 @@ module.exports = function (options, themeConfig) {
         var variation;
         var desktopScreenshot;
         var mobileScreenshot;
+        var version = themeConfig.getVersion();
 
         themeConfig.setVariation(variationIndex);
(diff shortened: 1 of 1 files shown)


server/plugins/stencil-editor/api/getVariations.js
```diff
@@ -68,7 +68,7 @@ module.exports = function (options, themeConfig) {
                 mobileScreenshot: mobileScreenshot,
                 demoUrl: variation.meta.demo_url,
                 documentationUrl: "string",
-                displayVersion: "string",
+                displayVersion: themeConfig.getVersion(),
```
bc-ong: On line 50, versionId has the same information.

mjschock: @bc-ong - i've updated it to share a variable.  i'm hesitant to hardcode versionId to something so i kept it as is but it looks like it might not be used.

bc-ong:
:+1:

PR #283, opened 2017-02-28
STENCIL-2930 Add support for google amp.

junedkazi:
Add support for google amp to the themes features supported list for validation.

@bigcommerce/stencil-team @mcampa @bookernath

### DIFF
--- a/lib/privateThemeConfig.schema.json
+++ b/lib/privateThemeConfig.schema.json
@@ -34,7 +34,7 @@
           "type": "array",
           "items": {
             "type": "string",
-            "enum": ["fully_responsive", "mega_navigation", "multi_tiered_sidebar_menu", "masonry_design", "frontpage_slideshow", "quick_add_to_cart", "switchable_product_view", "product_comparison_table", "complex_search_filtering", "customizable_product_selector", "cart_suggested_products", "free_customer_support", "free_theme_upgrades", "high_res_product_images", "product_filtering", "advanced_quick_view", "product_showcase", "persistent_cart", "one_page_check_out", "customized_checkout", "product_videos"]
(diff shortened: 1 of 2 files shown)


mcampa:
Hmm, I think we need to be more specific. I would say add a flag by page, `google_amp_product`, `google_amp_category`. Same for Theme Registry

junedkazi:
I think for now we should just leave it as `google_amp`. So if someone tried to prefix a page with `/amp/` & if it is something which is not supported it will still render the page as a regular resource. 

The problem will come up when we add support for a new page type let's say for example blog in the future & bcapp returns page type & template path for the route & if it does not exist in the theme it might throw a 500 on stapler.

So instead of adding new keys for each page type across cli, stencil, theme registry & bcapp we should build a feature detection system to look out for pages across amp directories. But it is beyond the scope of work for the big hack. 

We should create a new epic & continue to add new tickets/cleanup tickets based on issues we are aware of & the feedback we will receive from the beta to address all the issues before we open the feature to all merchants.

cc @mattolson @bookernath

mcampa:
I agree, we should have a "feature detection system". This could be solved by storing the manifest in the store config when the theme is installed

mattolson:
I'm thinking of something a little more flexible, where Theme Registry can inspect the theme for various traits and keep track of that data for later, passing it to bcapp when the theme is applied. This would be beyond the self-decleared "features" in config.json -- we'd augment that with information gathered from our own inspection of the theme. In the context of amp, that means which amp templates are included in the theme. This will be useful for checkout and for content injection as well.

bookernath:
I think the short-term fix Juned proposes is fine; but I do want to very quickly get to a place where we throw a 404 and not a 500 when we can't find the right template file.

For the first release of AMP the pages should only be discovered by the appropriate canonical tags in the parent non-AMP pages, so we're probably okay for the moment - in the spirt of limiting investment until AMP is proven.

PR #286, opened 2017-03-15
STENCIL-3243 run the theme bundling script (webpack) in a different process

mcampa:
[STENCIL-3243](https://jira.bigcommerce.com/browse/STENCIL-3243)

### Run the theme bundling script (webpack) in a different process (fork)

Webpack was run from the cli using a `require()` call to a file in the theme directory. This makes webpack to use dependencies from `stencil-cli` instead of only using dependencies from the theme itself

This requires a change in Cornerstone `stencil.conf.js` format, but is written to be backwards compatible.

Instead of `require`ing the file directly from `stencil-cli`, we are now using [`child_process.fork()`](https://nodejs.org/api/child_process.html#child_process_child_process_fork_modulepath_args_options) and communicating with the process over an IPC communication channel 

@bigcommerce/stencil-team


mcampa:
@lord2800

dstaley:
If this prevents Webpack from picking up JSPM's polyfill of `System.import`, I'm going to be incredibly happy. I think that's the only caveat I ran into when migrating to Webpack 2.

mcampa:
That is exactly what is doing @dstaley. We read your article, and we are in the process of upgrading Cornerstone to use Webpack2. So we thank you.

package.json
```diff
@@ -34,6 +34,7 @@
     "accept-language-parser": "^1.0.2",
     "archiver": "^0.14.4",
     "async": "^1.3.0",
+    "babel-eslint": "^7.1.1",
```
mjschock: thank you!  this gets the build working now, with the babel-eslint parser

test/_mocks/build-config/noready-config/stencil.conf.js
```diff
@@ -0,0 +1,52 @@
...
+        if (message === 'development') {
+            development();
+        }
+
+        if (message === 'production') {
+            production();
+        }
+    });
+
+    // process.send('ready');
```
mjschock: no need for the comment

mjschock:
LGTM; 👍   and i confirm that it works with the webpack 2 changes to Cornerstone

PR #298, opened 2017-04-28
Add support for Node 6 & 7

mcampa:
This uses the new `@bigcommerce/stencil-styles#1.0.6` which uses a [fork](https://github.com/bigcommerce-labs/node-sass) of `node-sass` compatible with node 6 & 7

Plus, adding AppVeyor for windows tests

@bigcommerce/stencil-team

### DIFF
--- a/lib/theme-config.spec.js
+++ b/lib/theme-config.spec.js
@@ -1,40 +1,42 @@
-var Code = require('code'),
-    Fs = require('fs'),
-    Hoek = require('hoek'),
-    Lab = require('lab'),
-    Path = require('path'),
-    Sinon = require('sinon'),
-    ThemeConfig = require('./theme-config'),
-    lab = exports.lab = Lab.script(),
-    themePath = Path.join(process.cwd(), 'test/_mocks/themes/valid'),
-    missingVariationsThemePath = Path.join(process.cwd(), 'test/_mocks/themes/missing-variation'),
-    bareBonesThemePath = Path.join(process.cwd(), 'test/_mocks/themes/bare-bones'),
-    describe = lab.describe,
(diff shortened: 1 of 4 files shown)


lib/theme-config.spec.js
```diff
@@ -1,40 +1,42 @@
...
-        it ('should allow overwriting of configPath, schemaPath, and variationName by calling getInstance with params', function(done) {
-            var secondConfigPath = '/fake/config.json';
-            var secondSchemaPath = '/fake/schema.json';
-            var secondVariationName = 'Second';
-            var secondThemeConfig;
-
-            secondThemeConfig = ThemeConfig.getInstance('/fake');
+        it ('should allow overwriting of configPath, schemaPath, and variationName by calling getInstance with params', done => {
+            const isWindows = Os.platform() === 'win32';
+            const secondConfigPath = isWindows ? '\\fake\\config.json' : '/fake/config.json';
```
mjschock: i think we use this [path.posix method]|(https://nodejs.org/api/path.html#path_path_posix) instead.  from [this page](https://nodejs.org/api/path.html#path_windows_vs_posix) i think we could do:

```
const normalize = path.posix.normalize;
expect(normalize(secondThemeConfig.configPath)).to.equal(normalize(secondConfigPath));
```

mcampa: awesome

mjschock:
LGTM 👍

lib/theme-config.spec.js
mcampa: doesn't work :( reverted

adammcarth:
Thanks @mcampa 👍

PR #302, opened 2017-05-04
STENCIL-3046 - Stencil Push

mjschock:
## WHAT
* implements `stencil push` command to bundle and upload a theme
* provides the option to apply the theme
* prompts for the deletion of existing private themes if the theme limit has been reached

screen shot 2017-06-22 at 5 02 03 pm

Theme Limit Reached GIF

## WHY
* [STENCIL-3046](https://jira.bigcommerce.com/browse/STENCIL-3046)

cc @bigcommerce/stencil-team

### DIFF
--- a/lib/stencil-push.js
+++ b/lib/stencil-push.js
@@ -1,0 +1,28 @@
+'use strict';
+const async = require('async');
+let utils = require('./stencil-push.utils');
+
+module.exports = stencilPush;
+
+function stencilPush(options, callback) {
+    options = options || {};
+    async.waterfall([
+        async.constant(options),
+        utils.readStencilConfigFile,
+        utils.getStoreHash,
+        utils.generateBundle,
+        utils.uploadBundle,
+        utils.notifyUserOfThemeLimitReachedIfNecessary,
+        utils.getThemesIfNecessary,
+        utils.promptUserToDeleteThemesIfNecessary,
+        utils.deleteThemesIfNecessary,
+        utils.uploadBundleAgainIfNecessary,
+        utils.notifyUserOfThemeUploadCompletion,
+        utils.pollForJobCompletion(),
+        utils.promptUserWhetherToApplyTheme,
+        utils.getVariations,
+        utils.promptUserForVariation,
+        utils.requestToApplyVariationWithRetrys(),
+        utils.notifyUserOfCompletion,
+    ], callback);
+}
--- a/lib/stencil-push.spec.js
+++ b/lib/stencil-push.spec.js
@@ -1,0 +1,91 @@
+'use strict';
+
+const Code = require('code');
+const Lab = require('lab');
+const lab = exports.lab = Lab.script();
+const describe = lab.describe;
+const expect = Code.expect;
+const it = lab.it;
+const sinon = require('sinon');
+const StencilPush = require('./stencil-push');
+const utils = require('./stencil-push.utils.js');
+const Wreck = require('wreck');
+const mockDb = {
+    _data: {},
+    get data() {
+        return this._data;
+    },
+    set data(data) {
+        this._data = data;
+    },
+}
+
+describe('stencil push', () => {
+    let sandbox;
+
+    lab.beforeEach(done => {
+        mockDb.data = {};
+        sandbox = sinon.sandbox.create();
+
+        sandbox.stub(Wreck, 'get').callsFake(requestStub);
+        sandbox.stub(Wreck, 'post').callsFake(requestStub);
+        sandbox.stub(Wreck, 'put').callsFake(requestStub);
+        sandbox.stub(utils, 'generateBundle').callsFake(utilStub({
+            bundleZipPath: 'bundleZipPath',
+        }));
+        sandbox.stub(utils, 'promptUserWhetherToApplyTheme').callsFake(utilStub({
+            applyTheme: true,
+        }));
+        sandbox.stub(utils, 'promptUserForVariation').callsFake(utilStub({
+            variationId: 'bold',
+        }));
+
+        function requestStub(url, options, callback) {
+            process.nextTick(() => {
+                callback(null, { statusCode: 200 }, mockDb.data);
+            });
+        }
+
+        function utilStub(data) {
+            return (options, callback) => {
+                process.nextTick(() => {
+                    callback(null, Object.assign({}, options, data));
+                });
+            };
+        }
+
+        done();
+    });
+
+    lab.afterEach(done => {
+        sandbox.restore();
+        done();
+    });
+
+    it('should throw an error if dotStencilFilePath is not provided', done => {
+        const throws = () => {
+            StencilPush();
+        };
+
+        expect(throws).to.throw('dotStencilFilePath is required!');
+        done();
+    });
+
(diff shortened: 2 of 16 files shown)


lib/stencil-push.js
```diff
@@ -0,0 +1,242 @@
...
+            }
+
+            callback(null, Object.assign({}, config, { storeHash: body.store_hash }));
+        });
+    });
+}
+
+internals.generateBundle = function(config, callback) {
+    const spawn = require('child_process').spawn;
+    const bundle = spawn(path.resolve(__dirname + '/../bin/stencil'), ['bundle']);
```
mcampa: Why do you need to spawn a process when you can call it directly?

mjschock: hmmm..  i didn't see a way to do that without duplicating the code from bin/stencil-bundle so i figured i might as well invoke the script.  but that makes me think, shouldn't we permit the options that are available to stencil bundle?  dest, name, and marketplace?

lib/stencil-push.js
```diff
@@ -0,0 +1,44 @@
...
+    options = options || {};
+    async.waterfall([
+        async.constant(options),
+        utils.readStencilConfigFile,
+        utils.getStoreHash,
+        utils.generateBundle,
+        // utils.generateUploadSignature,
+        utils.uploadBundle,
+        // utils.requestToProcessThemeUpload,
+        async.retryable({
```
mcampa: Can you move this to `stencil-push.utils`?

lib/theme-api-client.js
```diff
@@ -0,0 +1,176 @@
+'use strict';
+
+const api = 'https://api.integration.zone';
+// const api = 'https://api.bigcommerce.com';
```
mcampa: Can you uncomment this line and delete the other?
Please add a parameter to overwrite this url. Something like:
```
stencil push --host https://api.integration.zone
```

lib/stencil-push.spec.js
```diff
@@ -0,0 +1,94 @@
...
+
+const Code = require('code');
+const Lab = require('lab');
+const lab = exports.lab = Lab.script();
+const describe = lab.describe;
+const expect = Code.expect;
+const it = lab.it;
+const rewire = require("rewire");
+const sinon = require('sinon');
+const StencilPush = rewire('./stencil-push');
```
mcampa: hmm, didn't you find another way to mock without `rewire`?

lib/stencil-push.spec.js
```diff
@@ -0,0 +1,94 @@
...
+            "utils.generateBundle": utilStub({
+                bundleZipPath: 'bundleZipPath',
+            }),
+            "utils.promptUserWhetherToApplyTheme": utilStub({
+                applyTheme: true,
+            }),
+            "utils.promptUserForVariation": utilStub({
+                variationId: 'bold',
+            }),
+        });
```
mcampa: You don't need rewire to stub this. 
You can stub `utils` directly instead, something like this:

```js
const utils = require('./stencil-push.utils.js');
sinon.stub(utils, 'generateBundle');
```

lib/stencil-push.spec.js
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        done();
+    });
+
+    lab.afterEach(done => {
+        this.wreckGetStub.restore();
+        this.wreckPostStub.restore();
+        this.wreckPutStub.restore();
+        this.restoreStencilPush();
```
mcampa: Can you use `sandbox` instead? that way you don't need to restore all stubs, just call `sandbox.restore()`

https://github.com/bigcommerce/stencil-cli/blob/f3fe45612ae094d5d0dccc0cffb99acdf8959bd3/bin/stencil-bundle.spec.js#L43

lib/stencil-push.utils.js
```diff
@@ -0,0 +1,329 @@
...
+});
+
+module.exports = utils;
+
+function validateOptions(options, fields) {
+    options = options || {};
+    fields = fields || [];
+
+    fields.forEach(field => {
+        if (!_.has(options, field)) {
```
mcampa: why not just `!options.hasOwnProperty(field)`?

scripts/tdd
```diff
@@ -0,0 +1,20 @@
...
+            if (err) {
+                console.error(err);
+            }
+            if (stderr) {
+                console.error(stderr);
+            }
+            console.log(stdout);
+        });
+    });
+})
```
mcampa: Do you want to commit this file?

package.json
```diff
@@ -103,8 +105,10 @@
     "jschannel": "git@github.com:bigcommerce-labs/jschannel.git",
     "ng-common": "git@github.com:bigcommerce-labs/ng-common.git",
     "ng-stencil-editor": "git@github.com:bigcommerce-labs/ng-stencil-editor.git",
-    "stencil-preview-sdk": "git@github.com:bigcommerce-labs/stencil-preview-sdk.git",
+    "progress": "^2.0.0",
+    "request": "^2.81.0",
```
mcampa: Why are these two in `devDependencies`?

scripts/tdd
mjschock: nope - removed

package.json
mjschock: they've been moved

lib/theme-api-client.js
mjschock: this command ^ works now - and if you do `stencil push` it'll use the default prod endpoint.

lib/stencil-push.utils.js
mjschock: oh, this will allow nested fields, like so:

`options = validateOptions(options, ['config.normalStoreUrl']);`

lib/stencil-push.spec.js
mjschock: it's been removed, forgot about it

package.json
```diff
@@ -9,15 +9,16 @@
   "scripts": {
     "lint": "eslint .",
     "test": "npm run lint && npm run run-tests",
-    "run-tests": "lab bin/ server/ lib/ -t 80 -l -e 'development' -P spec",
+    "run-tests": "lab bin/ server/ lib/ -t 80 -l -C -e 'development' -P spec",
```
mcampa: what's the `-C` for?

mjschock: ha! it's for "color output" - it looked prettier when i was doing tdd.  i can remove it if you want

lib/theme-api-client.js
```diff
@@ -0,0 +1,174 @@
...
+        url: `${options.apiHost}/stores/${options.storeHash}/v3/themes`,
+    };
+
+    request(opts, (error, response, payload) => {
+        if (error) {
+            return callback(error);
+        }
+
+        if (response.statusCode !== 201) {
+            if (response.statusCode === 502 && payload && /upload limit/.test(payload.title)) {
```
mjschock: @junedkazi - can we change the statusCode to something more relevant or include a code within the body?  i'm worried that the `title` field's value may change in the future.  502 doesn't work here because it's used for other scenarios.

nickdengler:
LGTM 💚

Issue #306, opened 2017-05-19
npm WARN eslint-config-airbnb@6.2.0 requires a peer of eslint-plugin-react@^4.2.3 but none was installed.


wizardmadness:
node -v : v7.10.0

```
sudo npm install -g @bigcommerce/stencil-cli
Password:
npm WARN deprecated minimatch@0.3.0: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
npm WARN deprecated npmconf@2.1.2: this package has been reintegrated into npm and is now out of date with respect to npm
npm WARN deprecated minimatch@0.2.14: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
npm WARN deprecated graceful-fs@1.2.3: graceful-fs v3.0.0 and before will fail on node releases >= v7.0. Please update to graceful-fs@^4.0.0 as soon as possible. Use 'npm ls graceful-fs' to find it in the tree.
npm WARN deprecated minimatch@2.0.10: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
npm WARN deprecated node-uuid@1.4.8: Use uuid module instead
npm WARN deprecated good-reporter@3.1.0: This module is no longer being maintained and is no longer needed for custom Good reporters.
npm WARN deprecated es6-module-loader@0.16.6: This project has been deprecated for "npm install es-module-loader" based on the newer loader spec.
/usr/local/bin/stencil -> /usr/local/lib/node_modules/@bigcommerce/stencil-cli/bin/stencil
/usr/local/bin/stencil-start -> /usr/local/lib/node_modules/@bigcommerce/stencil-cli/bin/stencil-start
/usr/local/bin/stencil-init -> /usr/local/lib/node_modules/@bigcommerce/stencil-cli/bin/stencil-init

> @bigcommerce/node-sass@3.4.3 install /usr/local/lib/node_modules/@bigcommerce/stencil-cli/node_modules/@bigcommerce/node-sass
> node scripts/install.js

Binary downloaded and installed at /usr/local/lib/node_modules/@bigcommerce/stencil-cli/node_modules/@bigcommerce/node-sass/vendor/darwin-x64-51/binding.node

> @bigcommerce/node-sass@3.4.3 postinstall /usr/local/lib/node_modules/@bigcommerce/stencil-cli/node_modules/@bigcommerce/node-sass
> node scripts/build.js

[...]

PR #307, opened 2017-05-19
Remove eslint-config-airbnb dependency

mcampa:
This dependency is not used and is showing a warning to users

see https://github.com/bigcommerce/stencil-cli/issues/306


mcampa:
@bigcommerce/stencil-team

Issue #284, opened 2017-03-01
Browsersync issues with custom files


enkota:
Hey,

We are using a custom.scss file within the assets folder. The file is working as intended and compiling. It's appearing on the terminal that browser sync is noticing the changes however does not refresh the page for that file. Is there another step I am missing out here that I need to add new sass files to?

Thanks,

[BS] File event [change] : /Users/Bigcommerce/5ive1/stencil/assets/css/custom.css

mcampa:
Changes to scss files only refresh the stylesheets, not the entire page.

enkota:
@mcampa Thanks for the reply, yes I understand but the changes are not being made on the page when saving the style changes.

Bshally:
This is also happening to me, but is the case with all .scss files. The file event is registered in the terminal, but then none of the changes are injected into the browser and no notification pops up in the top right corner as it should to signal this happened.

enkota:
Yea, essentially any changes made within the custom.scss file (has been imported via theme.scss) will not trigger the live reloading of browsersync. As @Bshally has mentioned, the file event is registered in terminal but not refreshed on save.

mcampa:
investigating...

enkota:
Thanks @mcampa will test out in work on Tuesday!

PR #308, opened 2017-05-24
STENCIL-3480 Simplify and fix BrowserSync css reload issue

mcampa:
- Fixes: https://github.com/bigcommerce/stencil-cli/issues/284

Jira: https://jira.bigcommerce.com/browse/STENCIL-3480

@bigcommerce/stencil-team

### DIFF
--- a/bin/stencil-start
+++ b/bin/stencil-start
@@ -20,15 +20,11 @@ var Wreck = require('wreck');
 var jsonLint = require('../lib/json-lint');
 var versionCheck = require('../lib/version-check');
 var themePath = process.cwd();
-var allowedCssCompilers = ['scss', 'less'];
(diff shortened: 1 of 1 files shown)


willfaught:
Looking

willfaught:
Can you explain why these changes fix that problem?

mcampa:
We changed the css urls recently. So the way we were calling `BrowserSync.reload()` was wrong. And then I found that we don't need to pass the css file path because since version 2.6.0 we can use wildcards.

https://www.browsersync.io/docs/api#api-reload

Bshally:
Injection works now, thanks! I've created an issue for a bug occurring after this issue: [#309](https://github.com/bigcommerce/stencil-cli/issues/309)

Issue #318, opened 2017-07-07
"TypeError: callback is not a function" after upgrading to stencil 1.12.0


flyingL123:
I just upgraded to the latest version of stencil-cli, and now getting an error anytime I edit, or simply just open any template file. I upgraded from version 1.8 by running:

`npm install -g @bigcommerce/stencil-cli`

After the new version installed, I ran `stencil start`, and everything seemed to be working fine.

Then, I opened a template file to edit `vi templates/components/common/header.html`. As soon as the file opens, stencil-cli outputs the following error, and shuts down:

```
/Users/myusername/.nvm/versions/node/v4.2.1/lib/node_modules/@bigcommerce/stencil-cli/lib/template-assembler.js:26
            return callback(err);
                   ^

TypeError: callback is not a function
    at /Users/myusername/.nvm/versions/node/v4.2.1/lib/node_modules/@bigcommerce/stencil-cli/lib/template-assembler.js:26:20
    at /Users/myusername/.nvm/versions/node/v4.2.1/lib/node_modules/@bigcommerce/stencil-cli/lib/template-assembler.js:91:20
    at /Users/myusername/.nvm/versions/node/v4.2.1/lib/node_modules/@bigcommerce/stencil-cli/node_modules/async/dist/async.js:421:16
    at iteratorCallback (/Users/myusername/.nvm/versions/node/v4.2.1/lib/node_modules/@bigcommerce/stencil-cli/node_modules/async/dist/async.js:998:13)
    at /Users/myusername/.nvm/versions/node/v4.2.1/lib/node_modules/@bigcommerce/stencil-cli/node_modules/async/dist/async.js:906:16
    at ReadFileContext.callback (/Users/myusername/.nvm/versions/node/v4.2.1/lib/node_modules/@bigcommerce/stencil-cli/lib/template-assembler.js:112:24)
    at FSReqWrap.readFileAfterOpen [as oncomplete] (fs.js:303:13)
```

This occurs everytime a template file is openend or edited. It does not seem to occur when I edit a sass file. So far, I have only seen it when editing/opening/saving something in the `templates/` directory.

I am running node `v4.2.1` in case that matters.

Any idea why this is happening all of a sudden? Is there some sort of bug with the latest version?

flyingL123:
[...]

PR #319, opened 2017-07-07
Fix for issue #318

mjschock:
## WHAT
* this is a fix for #318 

## WHY
* the `assemble` method now requires a `templatesFolder` parameter (https://github.com/bigcommerce/stencil-cli/commit/49bcdc6fd79ecf09b70c080a8ea0ae1359d05c57#diff-4d9ab04ab5e144dcef049c51ef7a71fcR21)

cc @bigcommerce/stencil-team @flyingL123

### DIFF
--- a/bin/stencil-start
+++ b/bin/stencil-start
@@ -283,7 +283,7 @@ function assembleTemplates(templatePath, callback) {
             return file.replace(templatePath + Path.sep, '').replace('.html', '');
         });
 
-        Async.map(files, templateAssembler.assemble, function (err, results) {
+        Async.map(files, templateAssembler.assemble.bind(null, templatePath), function (err, results) {
             if (err) {
                 callback(err);
             }


mjschock:
@junedkazi - the commit message has been updated

mullinsr:
First reported on [StackOverflow](https://stackoverflow.com/questions/44962320/typeerror-callback-is-not-a-function-after-upgrading-to-stencil-1-12-0)

flyingL123:
Thanks everyone for the quick turnaround on this. I'm slightly unfamiliar with how this process works. At what point is this update available in the version I would get by running `npm install -g @bigcommerce/stencil-cli`? Is it there now or does something else need to occur?

PreciselyAlyss:
@flyingL123 It should be available through npm update/install. Are you still having issues with this?

flyingL123:
Nope, it's working now. Thanks for the follow-up.

PR #330, opened 2017-08-16
MERC-2617 Region manifest

mcampa:
#### What?

Stencil CLI needs to produce a list of regions within each of the template files.

Here I use regex to scan for region tags, building a full template per page to "guaranty" the order of the list returned.

I wrote a test case that considers "all" cases using `dynamicComponent` and `partial/block` helpers

@pedelman @mattolson @mjschock


mcampa:
Forgot to ping @tabayomi

lib/stencil-bundle.js
```diff
@@ -198,30 +221,31 @@ Bundle.prototype.getJspmBundleTask = function (jspmConfig) {
...
 
-        filePaths = filePaths.map((file) => {
-            return upath.toUnix(file.replace(this.templatesBasePath + Path.sep, '').replace('.html', ''));
+        const partials = filePaths.map(file => {
+            return Upath.toUnix(file.replace(this.templatesPath + Path.sep, '').replace('.html', ''));
         });
 
-        output.templates = filePaths;
-        this.manifest = output;
+        taskResults.manifest = {
```
mjschock: to keep it a bit cleaner, could we callback with taskResults instead?

lib/regions.js
```diff
@@ -0,0 +1,116 @@
...
+        return buildTemplateIncludes(templates, partial, visited);
+    });
+}
+
+/**
+ * Parse a template content and return an array of unique region objects
+ * @param {String} content
+ * @return {Array}
+ */
+function parseRegions(content) {
```
pedelman: Can we add some test coverage for the regex. Its hard to review whether this regex handles all the special cases.

mcampa: Can you provide the special cases?

lib/regions.spec.js
```diff
@@ -0,0 +1,97 @@
...
+            '{{{  region name="foobar1"  }}}': [{ name: 'foobar1' }],
+            '{{{  region  name=" "  }}}': [],
+            '{{{  region  name="invalid name"  }}}': [],
+            '{{  region  name="two_brackets"  }}': [],
+            '{{{  region   name=\'foobar\'  }}}': [{ name: 'foobar' }],
+            '{{{region name="foobar" type="widget"}}}': [{ name: 'foobar' }],
+            '{{{ region type="widget"  name="foobar"  }}}': [{ name: 'foobar' }],
+            '{{{ region name="foobar"  type="widget" }}}': [{ name: 'foobar' }],
+            '{{{ region name=\'foo\' }}} \n {{{ region name="bar" }}}': [{ name: 'foo' }, { name: 'bar' }],
+            '{{{region name=\'foo\'}}}{{{region name="bar"}}}{{{region name="foo"}}}': [{ name: 'foo' }, { name: 'bar' }],
```
mcampa: @pedelman let me know if this is enough

PR #348, opened 2017-11-28
adds hint text for dangling backslash

aisflat439:
#### What?

If you try to use `stencil push`  and your normalStoreUrl in the `.stencil` file has a trailing slash you get an unhelpful error. 

example:
`  "normalStoreUrl": "http://awesomeBCSandbox.mybigcommerce.com/",`

gives error 

`not ok -- StoreHashReadError: Response Error: 503 Service Unavailable`

#### Tickets / Documentation

none

none

### DIFF
--- a/bin/stencil-push
+++ b/bin/stencil-push
@@ -25,6 +25,9 @@ stencilPush(Object.assign({}, options, {
 }), (err, result) => {
     if (err) {
         console.log('not ok'.red + ` -- ${err}`);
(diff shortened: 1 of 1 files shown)


bookernath:
A better solution might be to drop the trailing slash when the file is saved coming out of `stencil init`? The thought being that preventing an error from occurring is better than notifying someone when it has.

aisflat439:
It's a much better solution. I just don't have time to implement an actual fix currently.

mjschock:
closing in favor of #369

PR #359, opened 2018-01-09
adding regex replace to remove .scss extension

NikitaPuza:
#### What?
Currently the css-assembler appends the .scss extension to the files its processing. This is fine in most cases since the default BigCommerce scss imports do not specify an extension. 

There are some frameworks/libraries that do specify the .scss extension such as UIkit(https://getuikit.com/docs/sass). 

If you attempt to import these scss files the cli will throw a file not found error. Removing the extension will resolve this but, you would need to do this for every import statement and every time the framework is updated. This creates extra friction when developing on stencil if you are integrating another framework.

#### Tickets / Documentation
Adding a simple replace regex to remove the .scss extension from an import before trying to find the file allows developers to use both.

Before:
screen shot 2018-01-09 at 12 56 02 pm

After:
screen shot 2018-01-09 at 12 57 28 pm

### DIFF
--- a/lib/css-assembler.js
+++ b/lib/css-assembler.js
@@ -44,7 +44,7 @@ function assemble(cssFiles, absolutePath, type, options, callback) {
     });
 
     function parseImports(cssFileAlias, callback) {
-        cssFileAlias = cssFileAlias.replace(/(\.scss)/g, '');
+        cssFileAlias = cssFileAlias.replace(/(\.scss)$/g, '');
         var file = Path.join(absolutePath, cssFileAlias + ext),
             cssFileAliasDir = Path.parse(cssFileAlias).dir;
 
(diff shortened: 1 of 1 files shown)


lib/css-assembler.js
```diff
@@ -44,6 +44,7 @@ function assemble(cssFiles, absolutePath, type, options, callback) {
     });
 
     function parseImports(cssFileAlias, callback) {
+        cssFileAlias = cssFileAlias.replace(/(\.scss)/g, '');
```
lord2800: It might be worth anchoring this to the end of the filename, i.e. `/(\.scss)$/g`.

NikitaPuza: Thats a good point. Updated in new commit.

Issue #360, opened 2018-01-10
Stencil push httpt error 404


rquiros:
### Expected behavior
Theme should be uploaded to store

### Actual behavior
Errors out with http error 404

[user@system tir]$  stencil push
not ok -- StoreHashReadError: Response Error: 404 Not Found
Please try again. If this error persists, please visit https://github.com/bigcommerce/stencil-cli/issues and submit an issue.

### Steps to reproduce behavior
Running stencil push command.

creativewhs:
I am getting this error too.
Did you find a solution?

trendyequine:
@creativewhs what template are you using? cornerstone, or a premium template?

creativewhs:
@trendyequine  It's the cornerstone template. Version 1.12.1

DIYVaporSupply:
Same
...\cornerstone>stencil push
not ok -- StoreHashReadError: Response Error: 404 Not Found
Please try again. If this error persists, please visit https://github.com/bigcom
merce/stencil-cli/issues and submit an issue.

petrkotek:
@DIYVaporSupply: what did you change in the `.stencil` file? The information look to be correct, but it still doesn't work. (The OAuth token does have modify permission `Themes:modify`)

```
{
  "normalStoreUrl": "https://mydomain.com/",
  "port": 3000,
  "clientId": "...id...",
  "accessToken": "...token...",
  "customLayouts": {
    "brand": {},
    "category": {},
    "page": {},
    "product": {}
  }
}
```

petrkotek:
can someone provide guidance? @junedkazi @mjschock //cc @aleachjr

petrkotek:
🎉  hurray, I figured out what was the problem.

The problem is a trailing slash in `normalStoreUrl` - there MUST BE NO TRAILING SLASH, only `"normalStoreUrl": "https://example.com"`.

Explanation: https://github.com/bigcommerce/stencil-cli/blob/08f6c9d8b061b239c543090abcccf2a927bb4419/lib/stencil-push.utils.js#L55 concatenates the `normalStoreUrl` with `/admin/oauth/info` and request to https://bcstore.com//admin/oauth/info` returns 404.

aleachjr:
Nice, thanks @petrkotek!

creativewhs:
Thanks @petrkotek, it works!

DIYVaporSupply:
@petrkotek
[...]

PR #369, opened 2018-02-21
STRF-4472 - Allow forward slash in store url

mjschock:
#### What?

* replaces a slash if present at the end of the store url so it's allowed and doesn't throw an error

#### Tickets / Documentation

- [STRF-4472](https://jira.bigcommerce.com/browse/STRF-4472)
- https://github.com/bigcommerce/stencil-cli/issues/360

#### Screenshots

* before
screen shot 2018-02-20 at 4 08 25 pm

* after
screen shot 2018-02-20 at 4 07 53 pm

@bigcommerce/storefront-team

### DIFF
--- a/lib/stencil-push.utils.js
+++ b/lib/stencil-push.utils.js
@@ -52,7 +52,7 @@ utils.readStencilConfigFile = (options, callback) => {
 utils.getStoreHash = (options, callback) => {
     options = validateOptions(options, ['config.normalStoreUrl']);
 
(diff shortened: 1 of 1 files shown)


bookernath:
@mjschock idea - why not be even more permissive and accept basically any URL regardless of path, query string, etc? I did this recently for archer - https://github.com/bigcommerce-labs/archer/pull/44

mjschock:
@bookernath 👍 i've made it more permissive, now all of these are okay:

http://example.com
https://example.com
example.com

http://example.com/
https://example.com/
example.com/

http://example.com/blah/
https://example.com/blah/
example.com/blah/

PR #372, opened 2018-03-29
Ability to locally save bundled theme after push (properly this time?)

willym1:
#### What?

When doing a normal stencil push (without specifying the filename of the bundle to upload), it first generates a bundle before pushing it to the store. The bundled file then gets thrown into a temp folder with no convenient way to retrieve it.

This commit adds a "save" option to save the bundled file into the current project folder. The current method is to do a stencil bundle first, then a stencil push with the name of the bundled file, but this skips those extra steps.

#### Screenshots

If no filename given, the bundle will have default filename:
example-1

Filename of bundle if a filename is given
example-2

### DIFF
--- a/lib/stencil-push.utils.js
+++ b/lib/stencil-push.utils.js
@@ -89,15 +89,23 @@ utils.getThemes = (options, callback) => {
 
 utils.generateBundle = (options, callback) => {
     let bundle;
+    let output = {
+        dest: os.tmpdir(),
+        name: uuid(),
+    };
 
     if (options.bundleZipPath) {
         return async.nextTick(callback.bind(null, null, options));
     }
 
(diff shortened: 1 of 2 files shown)


lib/stencil-push.utils.js
```diff
@@ -89,15 +89,23 @@ utils.getThemes = (options, callback) => {
...
     }
 
-    bundle = new Bundle(themePath, themeConfig, themeConfig.getRawConfig(), {
-        dest: os.tmpdir(),
-        name: uuid(),
-    });
+    if (options.saveBundleName) {
+        output = {
+            dest: themePath,
+            name: options.saveBundleName
```
junedkazi: You are missing a trailing comma here which is causing the build to fail

willym1: Fixed, seems good now

PR #385, opened 2018-05-24
Provide CDN URL to stencil-styles (compileCss)

jbruni:
#### What?

This PR is a "proof of concept" companion to https://github.com/bigcommerce/stencil-styles/issues/27

The proposed custom SASS function `stencilWebDAV` needs the base CDN URL, to be able to provide the WebDAV URL for assets.

Here, we are obtaining the CDN URL and passing it to the `compileCss` function defined at `stencil-styles`, so it can be used by `stencilWebDAV` custom SASS function. Just like already happens for the theme settings.

For local development using `stencil-cli`, this is not necessary, because an empty `cdnUrl` will result in a `/content` URL, exactly how the Handlebars `cdn` + `webdav` helper already does locally. The local server proxies the request, and the asset is returned in response.

This is why this PR is a "proof of concept". It is not necessary. But it fulfills the requirement of passing `cdnUrl` as an option to `compileCss`.

If it is easy to fulfill this requirement from client-side, it shall be even easier from server-side, where the CDN base URL must be readily available.

#### Tickets / Documentation

See  https://github.com/bigcommerce/stencil-styles/issues/27

### DIFF
--- a/server/plugins/theme-assets/theme-assets.module.js
+++ b/server/plugins/theme-assets/theme-assets.module.js
@@ -1,9 +1,15 @@
 const Boom = require('boom');
 const CssAssembler = require('../../../lib/css-assembler');
-const Utils = require('../../lib/utils');
+const Fs = require('fs');
 const Hoek = require('hoek');
+const jsonLint = require('../../../lib/json-lint');
 const Path = require('path');
+const Pkg = require('../../../package.json');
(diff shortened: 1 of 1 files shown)


junedkazi:
@jbruni I am going to close this one. Please reopen it if you would like to work on this one and see it through.

PR #388, opened 2018-05-30
Bundle all webpack files + release 1.15.0

mattolson:
#### What?
* Bundle all webpack config files. In Cornerstone 2.0, we split these up per environment, so we need to update the pattern used to select files for bundling.
* Bump version to 1.15.0 to prepare for release

#### Tickets / Documentation
- [Jira](https://jira.bigcommerce.com/browse/STRF-4971)
- [Cornerstone issue](https://github.com/bigcommerce/cornerstone/issues/1251)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@bigcommerce/stencil-cli",
-  "version": "1.14.3",
+  "version": "1.15.0",
   "description": "CLI tool to run BigCommerce Stores locally for theme development.",
   "main": "index.js",
   "engines": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bigcommerce/stencil-cli",
-  "version": "1.14.3",
+  "version": "1.15.0",
```
junedkazi: I thought we used `npm version` & `npm publish` to update versions.

mattolson: We could, but this basically does the same thing.

PR #412, opened 2018-10-01
Add node 10 to .travis.yml

bookernath:
#### What?

Adding Node 10 to the versions travis checks.

For now, this is mostly just to see what tests fail so we can start working on Node 10 support.


mattolson:
We need to build our custom fork of node-sass for Node 10.x first.

PR #441, opened 2019-02-26
Docs: upgraded paper

bc-zoharmuzafi:
#### What?
Bump paper version to 2.0.11


mattolson:
We should add package-lock.json as well for this repo. Can you do that as part of this PR?

PR #448, opened 2019-02-27
fix: added npm install to gulp task

bc-zoharmuzafi:
#### What?
After starting to check in package-lock.json, we should adjust the gulp task for releasing to run npm install prior the push of the release to github.

#### Screenshots
screen shot 2019-02-27 at 2 20 06 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,3 @@
-<a name="1.15.6"></a>
-## <small>1.15.6 (2019-02-26)</small>
-
-* docs: updated paper to 2.0.11 + checked package-lock (#446) ([54ddad0](https://github.com/bigcommerce/stencil-cli/commit/54ddad0))
-* fix: revert paper_2.0.11 pr (#445) ([adf424f](https://github.com/bigcommerce/stencil-cli/commit/adf424f))
-* Docs: upgraded paper (#441) ([26b3937](https://github.com/bigcommerce/stencil-cli/commit/26b3937))
-* feat(theme): add csrf protection to the list of valid features ([fe4b795](https://github.com/bigcommerce/stencil-cli/commit/fe4b795))
-
-
-
 <a name="1.15.5"></a>
 ## <small>1.15.5 (2019-02-05)</small>
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -10,6 +10,7 @@ const git = require('gulp-git-streamed');
 const gulp = require('gulp');
 const gulpif = require('gulp-if');
 const gutil = require('gulp-util');
+const supportedLockFileVersion = [1];
 const prompt = require('gulp-prompt');
 const semver = require('semver');
 
@@ -74,7 +75,13 @@ function bumpTask() {
             targetVersion = (res.type !== 'custom') ? semver.inc(currentVersion, res.type) : res['custom-version'];
             responses = res;
 
-            return gulp.src('package.json')
+            const packageLock = require('./package-lock.json');
+
+            if (!supportedLockFileVersion.includes(packageLock["lockfileVersion"])) {
+                throw new Error(`Release script only supports version ${supportedLockFileVersion}`);
+            }
+
+            return gulp.src(['package.json', 'package-lock.json'])
                 // bump the version number in those files
                 .pipe(bump({ version: targetVersion }))
                 // save it back to filesystem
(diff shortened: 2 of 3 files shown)


gulpfile.js
```diff
@@ -93,14 +93,20 @@ function deployWebpack() {
         .on('error', logError);
 }
 
+function installNpmDependencies() {
+    return gulp.src('package.json')
+        .pipe(exec('npm install'), logError)
```
lord2800: Does it make sense to check if the version of npm supports `ci` and if there's also a package-lock.json file, to use `npm ci` instead?

mattolson: If we use `npm install` at build time, we run the risk of making minor version changes at build time that were not validated by the author.

If we can't use `npm ci` to update the package version in `package-lock.json` (which would ensure that only the version number is updated and no dependencies), then I think we should modify the bump task to bump the version number in both `package.json` and `package-lock.json`.

gulpfile.js
```diff
@@ -87,6 +91,21 @@ function bumpTask() {
         .on('error', logError);
 }
 
+function bumpPackageLock(targetVersion) {
+    if (!lockfileVersions.includes(packageLock["lockfileVersion"])) {
+        throw new Error("check lockfileVersions in package-lock.json");
```
mattolson: Should say `Release script only supports version ${supportedLockFileVersion}`

gulpfile.js
```diff
@@ -10,6 +10,8 @@ const git = require('gulp-git-streamed');
 const gulp = require('gulp');
 const gulpif = require('gulp-if');
 const gutil = require('gulp-util');
+const lockfileVersions = [1];
```
mattolson: How about `supportedLockFileVersion`, and make it a single value rather than a list. If we get to the point that we need to support a new version, we'll likely need to re-evaluate this code anyway.

gulpfile.js
```diff
@@ -10,6 +10,8 @@ const git = require('gulp-git-streamed');
...
+const packageLock = require('./package-lock.json');
```
mattolson: I would remove this and instead rely on `gulp.src`

gulpfile.js
```diff
@@ -87,6 +91,21 @@ function bumpTask() {
...
+        throw new Error("check lockfileVersions in package-lock.json");
+    }
+    
+    return gulp.src('package-lock.json')
+        // bump the version number in those files
+        .pipe(bump({ version: targetVersion }))
+        // save it back to filesystem
+        .pipe(gulp.dest(process.cwd()))
+        // change last modified date
+        .pipe(exec('touch -c package.json'))
```
mattolson: Should be `package-lock.json`

gulpfile.js
```diff
@@ -74,6 +76,8 @@ function bumpTask() {
             targetVersion = (res.type !== 'custom') ? semver.inc(currentVersion, res.type) : res['custom-version'];
             responses = res;
 
+            bumpPackageLock(targetVersion);
```
mattolson: There is a simpler approach to this. Look at the readme for gulp-bump: https://github.com/stevelacy/gulp-bump#bumping-version-and-outputting-different-files

gulpfile.js
bc-zoharmuzafi: @mattolson we will need that outside the scope that use the src, to validate the lockfileVersion

CHANGELOG.md
```diff
@@ -1,13 +1,3 @@
-<a name="1.15.6"></a>
```
lord2800: Is this change intended?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bigcommerce/stencil-cli",
-  "version": "1.15.6",
```
lord2800: Is this change intended?

gulpfile.js
lord2800: Do we need to read it in up here, or can we do that in `bumpTask`?

CHANGELOG.md
bc-zoharmuzafi: @lord2800 yes, this is part or "rolling back" from the release that we made and re-release 1.15.6

package.json
bc-zoharmuzafi: @lord2800 same as the CHANGELOG

gulpfile.js
bc-zoharmuzafi: @lord2800 Can do it either way, we use it only in bumpTask anyway. I can change that if @mattolson will accept my previous comment related to not removing that completely.

CHANGELOG.md
lord2800: Got it.

gulpfile.js
```diff
@@ -10,6 +10,7 @@ const git = require('gulp-git-streamed');
 const gulp = require('gulp');
 const gulpif = require('gulp-if');
 const gutil = require('gulp-util');
+const supportedLockFileVersion = [1];
```
lord2800: 🍹 Since this is an array, it should probably be called `supportedLockFileVersions`.

PR #470, opened 2019-05-22
STRF-6383 Update stencil bundle to check size for generated parsed parsed/templates

bc-williamkwon:
#### What?
Add checks for template size.  Refactored bundleParsedFiles function to return an array.  The contents of this array are determined by checking the size of the data that is being added to a parsed template file and if the size is greater than a megabyte which is determined by the length of the stringified data as 1 byte is 1 char then the filename is added to the templatesArr.  Also refactored bundleTaskRunner to accept the returned array from bundleParsedFiles and check to see if the length of the array is greater than 0  if it is then we console.error and list the failed filenames.

#### Tickets / Documentation

### DIFF
--- a/lib/stencil-bundle.js
+++ b/lib/stencil-bundle.js
@@ -1,6 +1,6 @@
 'use strict';
-
-const MAX_SIZE_BUNDLE = 1048576 * 50; //50MB
+const MEGABYTE=1048576;
+const MAX_SIZE_BUNDLE = MEGABYTE * 50; //50MB
 const PATHS_TO_ZIP = [
     'assets/**/*',
     '!assets/cdn/**',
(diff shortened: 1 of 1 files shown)


lib/stencil-bundle.js
```diff
@@ -308,16 +308,19 @@ function bundleTaskRunner(callback) {
             bundleThemeFiles(archive, this.themePath, this.configuration);
 
             // zip all generated files
-            bundleParsedFiles(archive, taskResults);
-
+            const failedTemplatesArr=bundleParsedFiles(archive, taskResults);
```
megdesko: 🍹 spaces on either side of `=`

lib/stencil-bundle.js
```diff
@@ -354,14 +357,16 @@ function bundleThemeFiles(archive, themePath, configuration) {
...
  * @param {Archive} archive
  * @param {Object} taskResults
+ * @returns {Array} 
  */
 function bundleParsedFiles(archive, taskResults) {
     const archiveJsonFile = (data, name) => {
         archive.append(JSON.stringify(data), { name });
     }
-
+    const templatesArr=[];
```
megdesko: should this be a `var` since it can be modified in this method?

megdesko: Unrelated to this: if the idea is that we might (at a later date) have a series of errors to pass back, maybe this should be called `errorsArr` or something along those lines?

bc-williamkwon: No to first as const creates a read only reference to a value.  We aren't changing what it refers to.  The second one may be a good idea.

carsonreinke: @megdesko You mean `let` ;)

PR #471, opened 2019-05-22
Improved theme validation

bc-williamkwon:
#### What?

Stencil CLI: Fail stencil bundle if we fail to generate dist files
Update stencil bundle to check size for generated parsed parsed/templates

#### Tickets / Documentation

https://jira.bigcommerce.com/browse/STRF-6529


lib/stencil-bundle.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
-
-const MAX_SIZE_BUNDLE = 1048576 * 50; //50MB
+const MEGABYTE=1048576;
+const MAX_SIZE_BUNDLE = MEGABYTE * 50; //50MB
```
mattolson: Probably don't need that comment anymore :D

lib/stencil-bundle.js
```diff
@@ -276,8 +276,26 @@ function validateTheme(callback) {
...
+ * @returns {Boolean}
+ * 
+ */
+Bundle.prototype.distCheck = function () {
+    const destinationFolder = Path.join(this.themePath,'/assets/dist');
+    if (!Fs.existsSync(destinationFolder)) {
+        return false;
+    } else {
+        Readdir(destinationFolder, function(err,files) {
+            if (err || !Fs.existsSync(Path.join(destinationFolder,'theme-bundle.main.js')) || files.length <= 3) {
```
mattolson: This seems a bit magical to me. Why 3 files? Are there specific files that should always be there?

lib/stencil-bundle.js
```diff
@@ -288,6 +306,10 @@ function bundleTaskRunner(callback) {
         defaultName = 'Theme.zip';
     }
 
+    const distExists = this.distCheck();
```
mattolson: I'm surprised that the dist files exist at this point in time. They shouldn't be generated until the tasks are run below, specifically `tasks.theme`. Instead of putting this here, you should put it inside the callback for `Async.parallel` below.

lib/stencil-bundle.js
```diff
@@ -354,14 +379,16 @@ function bundleThemeFiles(archive, themePath, configuration) {
  * Archive all generated files (ex. parsed files)
  * @param {Archive} archive
  * @param {Object} taskResults
+ * @returns {Array} 
  */
-function bundleParsedFiles(archive, taskResults) {
+Bundle.prototype.bundleParsedFiles=function (archive, taskResults) {
```
mattolson: There shouldn't be any need to make this part of the Bundle object. The changes you're making are purely functional and don't need access to the Bundle object.

lib/stencil-bundle.js
bc-williamkwon: report.html, main and main.map.  i was expecting there to at least be a bundle-chunk file.

lib/stencil-bundle.js
bc-williamkwon: That change was in order to be able to unit test it.

lib/stencil-bundle.js
mattolson: Ok, maybe this should be a list of expected files and a loop to check for them. That would make the code clearer.

lib/stencil-bundle.js
mattolson: Ok, I recommend splitting this change out into a separate PR so we can discuss other potential strategies for that. It's also an unrelated change, and it's best to do them separately.

lib/stencil-bundle.js
bc-williamkwon: So the main issue is that the dist folder itself is not being created.  So instead of doing the stringent checking here, I am going to just check for the dist path.

lib/stencil-bundle.js
bc-williamkwon: Used a fresh copy of stencil-cli and made changes to reflect only the dist folder check.  Will make another pull request with the template check after version updating on the font-display stuff.

lib/stencil-bundle.js
```diff
@@ -292,6 +302,11 @@ function bundleTaskRunner(callback) {
         if (err) {
             return callback(err);
         }
+        const distExists = this.distFolderCheck();
+
+        if (!distExists) {
```
mattolson: What about `options.dest`?

lib/stencil-bundle.js
```diff
@@ -278,6 +278,16 @@ function validateTheme(callback) {
     });
 }
 
+/**
+ * check for dist folder
+ * @param {*}
+ * @returns {Boolean}
+ * 
+ */
+Bundle.prototype.distFolderCheck = function() {
```
mattolson: Is it enough to only check for the folder existence? Do we remove this folder on every run?

bc-williamkwon: The issue was that the dist folder was not being generated so I thought this should be enough.  The folder is not removed on every run.

lib/stencil-bundle.js
bc-williamkwon: this.options.dest is undefined for me.

lib/stencil-bundle.js
bc-williamkwon: I can put back the check for files but a for loop check would be difficult as bundle chunk files may go from 2 to 4 or not start off at 0(?) and also a check for the main file.

lib/stencil-bundle.js
bc-williamkwon: I just realized this.options.dest is an option for outputFolder which this.themePath can also substitute for.  For distExists to be true, this.themePath has to exist.  I can be more explicit about that though if you would like.

mattolson: If `options.dest` is provided, then the check in `distFolderCheck` will not work because it assumes `this.themePath`.

lib/stencil-bundle.js
mattolson: If the dist folder is not removed on each run, then this check will always return true after the first run, even if the files are not generated properly.

PR #475, opened 2019-05-27
fix: strf-6383 add check for template size

bc-williamkwon:
#### What?

Description
We need to validate the size of generated parsed/templates and make sure each of them are under 1MB.

We should fail the bundle process and notify on the terminal which generated files are greater than 1MB.

We are seeing a bunch of cases where zip files are under 50MB but the uploads still fail due to generated parsed/templates files greater than 1MB.

#### Tickets / Documentation


junedkazi:
No changelog for cli it is auto generated on release.

PR #478, opened 2019-05-31
Remove Node 6 from Travis

bookernath:
Node 6 is EOL

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ addons:
   ssh_known_hosts:
     - github.com 
 node_js:
-  - 6
   - 8
 cache:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -6,7 +6,6 @@ addons:
   ssh_known_hosts:
     - github.com 
 node_js:
-  - 6
```
junedkazi: We should also update `appveyor.yml`

junedkazi: We should also update the engine in `package.json` and regenerate package lock file.

bookernath:
@junedkazi ♻️

PR #496, opened 2019-07-15
feat: update stencil-paper to add getImageSrcset helper

bookernath:
#### What?

Bringing in latest stencil-paper so that new a new helper (`getImageSrcset`) can be used for local dev.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,14 +39,15 @@
       }
     },
     "@bigcommerce/stencil-paper": {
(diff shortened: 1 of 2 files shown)


bc-williamkwon:
@bigcommerce/storefront-team Can't approve due to not having ability but not sure about loss of two dependencies.

bookernath:
@junedkazi thanks! Feel free to merge and do the npm release when you get a sec 👍

PR #525, opened 2019-10-10
Adding Support for Node 10

ncheikh:
#### What?

Updating stencil styles dependency which allows node 10 support.

#### Tested
- [x] Locally against cornerstone & node 8
- [x] Locally against cornerstone & node 10

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,10 @@ addons:
     - github.com 
 node_js:
   - 8
+  - 9
+  - 10
+  - 11
+
 cache:
   directories:
     - node_modules
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,7 @@ addons:
     - github.com 
 node_js:
   - 8
-  - 9
   - 10
-  - 11
-  - 12
 
 cache:
   directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,8 @@ node_js:
   - 8
   - 9
   - 10
+  - 11
+  - 12
 
 cache:
   directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ node_js:
   - 8
   - 9
   - 10
-  - 11
 
 cache:
   directories:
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "CLI tool to run BigCommerce Stores locally for theme development.",
   "main": "index.js",
   "engines": {
-    "node": ">= 8.0.0 <12.0.0"
+    "node": ">= 8.0.0 <11.0.0"
```
junedkazi: This should be `<12` to support `11.x`

xzyfer:
>Waiting on: dlmanning/gulp-sass#748 for node 12

No need to wait. Node Sass already supports node 12.

.travis.yml
```diff
@@ -7,6 +7,8 @@ addons:
     - github.com 
 node_js:
   - 8
+  - 10
```
junedkazi: Should we add 9 as well ?

package.json
```diff
@@ -4,7 +4,7 @@
   "description": "CLI tool to run BigCommerce Stores locally for theme development.",
   "main": "index.js",
   "engines": {
-    "node": ">= 8.0.0 <9.0.0"
+    "node": ">= 8.0.0 <11.0.0"
```
junedkazi: Also can we also update the readme to reflect node version support.

ncheikh: Good catch!

.travis.yml
ncheikh: I don't think so, odd versions are only supported for a short time and I don't think we need to support node versions that aren't supported for cli.  In the future I can see doing the leading version like 13 for early adopters and it helps us make sure we are in a good position for the next even/lts release.

Thoughts?

@junedkazi @bookernath @bc-williamkwon 

https://nodejs.org/en/about/releases/

README.md
```diff
@@ -6,7 +6,7 @@
 The BigCommerce server emulator for local theme development.
 
 ## Install
-_Note: Stencil requires the Node.js runtime environment, version 8.x. We do not yet have support for Node 10 or greater._
+_Note: Stencil requires the Node.js runtime environment, version 8.x or 10.x We do not yet have support for Node 12 or greater._
```
ncheikh: @junedkazi @bookernath Let me know if we want different messaging here.

PR #534, opened 2019-11-21
Add error messages only if payload data exists

kylebolstad:
#### What?

When a job fails with a status code not equal to 200, its payload data doesn't exist. Error messages now only get set when payload data exists.


bookernath:
@kylebolstad thanks for your contribution! Do you mind rebasing and fixing any test failures?

junedkazi:
@kylebolstad I am going to close this one. Please reopen it if you would like to work on this one and see it through.

PR #539, opened 2019-12-12
feat(translate): added the schemaTranslations.json file into a bundle

RomanKrasinskyi:
#### What?

Added the schema Translations.json into a bundle for translate schema.json labels. It will allow developers to create a theme with translations for labels from schema.json which are displayed in store-design

#### Tickets / Documentation

- [Ticket](https://jira.bigcommerce.com/browse/MERC-5786)

Screen Shot 2019-12-12 at 5 41 59 PM


RomanKrasinskyi:
@pedelman

bookernath:
Hi @RomanKrasinskyi can you show an example of what the .json file might look like?

RomanKrasinskyi:
@bookernath schemaTranslations.json looks like:
```
{
    'i18n.backgroundAndLines': {
        en: 'Background and lines',
        uk: 'Фон і лінії',
    },
    'i18n.bannerBackground': {
        en: 'Banner background',
        uk: 'Банер фону',
    },
    'i18n.buttonsAndIcons': {
        en: 'Buttons & Icons',
        uk: 'Кнопки та значки',
    },
    'i18n.checkoutPage': {
        en: 'Checkout Page',
        uk: 'Сторінка оформлення замовлення',
    },
    'i18n.global': {
        en: 'Global',
        uk: 'Глобальні',
    },
    'i18n.headerAndFooter': {
        en: 'Header & Footer',
        uk: 'Заголовок і колонтитул',
    },
    'i18n.homePage': {
        en: 'Home Page',
        uk: 'Домашня сторінка',
    },
    'i18n.paymentButtons': {
        en: 'Payment Buttons',
        uk: 'Кнопки оплати',
    },
    'i18n.products': {
        en: 'Products',
        uk: 'Продукція',
    },
}
```
In `schema.json` we can use keys such as `i18n.products` for labels which should translatable

pedelman:
@RomanKrasinskyi do we plan to add validation for the schemaTranslations object in this PR or in a follow up? It would be pretty straightforward to add a json schema to make sure the translations have the expected structure.

RomanKrasinskyi:
@pedelman Ok, I will add validation in this PR.

RomanKrasinskyi:
@pedelman I've added validator for the schemaTranslations.json. Could you please review?

pedelman:
@RomanKrasinskyi I have been playing around with this branch locally, really nice job on validations. I found a few edge cases we may want to consider.

1. When schemaTranslations.json file has empty contents. When initially running through this code, I created a new blank file and ran stencil bundle. I got the following error. I wonder if its possible to give a more user friendly message in this case.

```
$ touch schemaTranslations.json
$ node ~/dev/stencil-cli/bin/stencil bundle
Validating theme...
/Users/patrick.edelman/dev/stencil-cli/lib/json-lint.js:10
            throw new Error(file + ' - ' + e.message);
            ^

Error: /Users/patrick.edelman/dev/cornerstone/schemaTranslations.json - Parse error on line 1:

^
Expecting 'STRING', 'NUMBER', 'NULL', 'TRUE', 'FALSE', '{', '[', got 'EOF'
    at Object.parse (/Users/patrick.edelman/dev/stencil-cli/lib/json-lint.js:10:19)
    at ThemeConfig.getRawSchemaTranslations (/Users/patrick.edelman/dev/stencil-cli/lib/theme-config.js:436:21)
    at BundleValidator.validateSchemaTranslations (/Users/patrick.edelman/dev/stencil-cli/lib/bundle-validator.js:194:41)
    at /Users/patrick.edelman/dev/stencil-cli/node_modules/async/dist/async.js:3880:24
    at replenish (/Users/patrick.edelman/dev/stencil-cli/node_modules/async/dist/async.js:1011:17)
    at /Users/patrick.edelman/dev/stencil-cli/node_modules/async/dist/async.js:1016:9
    at eachOfLimit (/Users/patrick.edelman/dev/stencil-cli/node_modules/async/dist/async.js:1041:24)
    at /Users/patrick.edelman/dev/stencil-cli/node_modules/async/dist/async.js:1046:16
    at _parallel (/Users/patrick.edelman/dev/stencil-cli/node_modules/async/dist/async.js:3879:5)
    at Object.series (/Users/patrick.edelman/dev/stencil-cli/node_modules/async/dist/async.js:4735:5)
```

2. When the value of the translations is empty string. Take the following schema as example. Maybe we can give an error that we should not allow empty string as the value.

```json
{
  "i18n.headings.global": {
    "default": ""
  }
}
```

3. When we have schemaTranslations which are not found in the schema. Currently we are tracking when we have a value added to schema.json, it must be present in the schemaTranslations, maybe we can also check opposite case as well (might be useful to catch typos).

lib/bundle-validator.js
```diff
@@ -106,6 +107,119 @@ function validateThemeConfiguration(callback) {
     callback(null, true);
 }
 
+/**
+ * Find translatable strings
+ * @param {object} schema
+ * @param {function} callback
+ */
+function getTranslatableStrings(schema, callback) {
+    var trackedKeys = ['name', 'content', 'label', 'settings', 'options'];
```
pedelman: I think we may also want to track the `group` option.

https://github.com/bigcommerce/cornerstone/blob/master/schema.json#L54

pedelman: I think we should use const instead of var if possible.

lib/bundle-validator.js
```diff
@@ -106,6 +107,119 @@ function validateThemeConfiguration(callback) {
...
+
+    schema.forEach(element => {
+        Object.keys(element).forEach(key => {
```
pedelman: I prefer to use Object.entries, can you see if this is available to use?

```
Object.entries(element).forEach([key, value] => {
   ...
}

```

lib/bundle-validator.js
```diff
@@ -106,6 +107,119 @@ function validateThemeConfiguration(callback) {
...
+            const value = element[key];
+
+            if (trackedKeys.indexOf(key) === -1) {
```
pedelman: I think includes is a bit easier to read / follow.

```
trackedKeys.includes(key)
```

lib/bundle-validator.js
```diff
@@ -106,6 +107,119 @@ function validateThemeConfiguration(callback) {
...
+ * @param {object} schema
+ * @return {array}
+ */
+function getI18nKeys(schema) {
+    var keys = [];
+
+    getTranslatableStrings(schema, (value) => {
+        if (
+            value
+            && keys.indexOf(value) === -1
```
pedelman: Another place we can consider to use `includes`

lib/bundle-validator.js
```diff
@@ -106,6 +107,119 @@ function validateThemeConfiguration(callback) {
...
+    });
+}
+
+/**
+ * Find i18n keys in schema
+ * @param {object} schema
+ * @return {array}
+ */
+function getI18nKeys(schema) {
+    var keys = [];
```
pedelman: Lets review all usage of var in this PR and see if we can use const in each case.

lib/bundle-validator.js
```diff
@@ -106,6 +107,119 @@ function validateThemeConfiguration(callback) {
...
+ * Find missed i18n keys in schemaTranslations
+ * @param {array} keys
+ * @param {object} translations
+ * @return {array}
+ */
+function findMissedKeys(keys, translations) {
+    var translationsKeys = Object.keys(translations);
+    var missingKeys = [];
+
+    keys.forEach(key => {
```
pedelman: I think we may be able to refactor this into just using a `.filter`.

`Object.keys(translations).filter(key => translationsKeys.includes(key));`

lib/bundle-validator.js
```diff
@@ -106,6 +107,119 @@ function validateThemeConfiguration(callback) {
...
+
+    if ((validation.errors && validation.errors.length > 0) || (missedKeys && missedKeys.length > 0)) {
+        errorMessage = 'Your theme\'s schemaTranslations.json has errors:'.red;
+
+        missedKeys.forEach(key => {
+            errorMessage += '\r\nmissing translation key "'.red + key.red + '"'.red;
+        });
+
+        validation.errors.forEach(error => {
+            errorMessage += '\r\nschemaTranslations'.red + error.stack.substring(8).red;
```
pedelman: What is the substring call needed for?

lib/bundle-validator.js
```diff
@@ -106,6 +107,119 @@ function validateThemeConfiguration(callback) {
     callback(null, true);
 }
 
+/**
+ * Find translatable strings
+ * @param {object} schema
+ * @param {function} callback
+ */
+function getTranslatableStrings(schema, callback) {
```
pedelman: What are your thoughts to keep some of these helper methods inside a subdirectory, this file can be about the process of running validations, less about the implementation details.

lib/schemaTranslations.schema.json
```diff
@@ -0,0 +1,22 @@
+{
```
pedelman: 👏

lib/bundle-validator.js
```diff
@@ -106,6 +107,119 @@ function validateThemeConfiguration(callback) {
...
+}
+
+/**
+ * Ensure that schema translations exists and there are no missing keys.
+ * @param {function} callback
+ * @return {function} callback
+ */
+function validateSchemaTranslations(callback) {
+    var v = new Validator();
+    var validatorTranslations = './schemaTranslations.schema.json';
```
jkanive: Does this have to be `schemaTranslations.schema.json`? Can we have the filename as just `schemaTranslations.json` and move it under `/lib/validator` directory?

lib/bundle-validator.js
RomanKrasinskyi: Could you please provide more information on the use of the `group` option? I didn't find any usage.

lib/bundle-validator.js
RomanKrasinskyi: `error.stack` is a string with some text like this: `instance is not of a type(s) object`.
Method substring returns ` is not of a type(s) object`, so we can make an error message like this: `schemaTranslations is not of a type(s) object`

lib/bundle-validator.js
RomanKrasinskyi: Ok. I will do

lib/bundle-validator.js
RomanKrasinskyi: Ok. I will do it

lib/schemaTranslations.schema.json
```diff
@@ -0,0 +1,22 @@
...
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
junedkazi: I don't believe the existing json schema package which we use supports draft 7.

lib/bundle-validator.js
pedelman: I can find examples in cornerstone schema, but it seems like there is a bug where store design is not displaying the grouped selects. Let me bring this back to the team and see what the desired state should be.

lib/schemaTranslations.schema.json
pedelman: Lets make sure this schema is the same as in this theme-registry pr: https://github.com/bigcommerce/theme-registry/pull/1016

lib/bundle-validator.js
RomanKrasinskyi: I added `group` options into `trackedKeys`

lib/schemaTranslations.schema.json
RomanKrasinskyi: @junedkazi Yeah, I checked. `jsonschema` package supports draft 4.
Thanks, I will fix it.

RomanKrasinskyi:
@pedelman I've updated validator, please check

pedelman:
@RomanKrasinskyi looks like there is minor conflict, lmk when this is fixed up and ill quickly approve again 👍

junedkazi:
Thank you @RomanKrasinskyi @pedelman for all the work which went into this one.

PR #541, opened 2019-12-14
fix: fix origin sent with graphql requests

bookernath:
#### What?

The JWT auth used for [GraphQL Storefront API](https://developer.bigcommerce.com/api-docs/storefront/graphql/graphql-storefront-api-overview) relies on the `origin` header sent by the browser to understand if the request should be allowed. For CLI, this origin was being sent as `localhost`, which was being rejected.

This PR uses CLI's proxying capabilities to overwrite/forge the correct `origin` header for the request so that GraphQL works again.

### DIFF
--- a/bin/stencil-init.spec.js
+++ b/bin/stencil-init.spec.js
@@ -14,7 +14,7 @@ describe('stencil init', () => {
     let sandbox;
 
     lab.beforeEach(done => {
-        sandbox = Sinon.sandbox.create();
+        sandbox = Sinon.createSandbox();
         sandbox.stub(console, 'log');
         sandbox.stub(console, 'error');
         done();
--- a/server/plugins/router/router.module.js
+++ b/server/plugins/router/router.module.js
@@ -16,6 +16,7 @@ const internals = {
         favicon: '/favicon.ico',
         stencilEditor: '/stencil-editor',
         updateParam: '/stencil-editor/update-param',
+        graphQL: '/graphql',
     },
 };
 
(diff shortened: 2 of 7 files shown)


server/plugins/router/router.module.js
```diff
@@ -130,6 +131,33 @@ internals.registerRoutes = function(server, next) {
...
+        {
+            method: ['GET', 'POST'],
+            path: internals.paths.graphQL,
+            handler: {
+                proxy: {
+                    mapUri: function (req, cb) {
+                        return cb(
+                            null,
+                            `${internals.options.storeUrl}${req.path}`,
+                            Object.assign(
```
bookernath: I originally had this as a new arrow function with a spread operator, but this made tests fail due to a bug in [eslint/espree](https://github.com/eslint/espree/pull/426) which was only fixed a few months ago. Upgrading our version of espree by 4 major versions seemed prohibitively complicated to make this one change, so I think we should just stick with the weird syntax for now.

server/plugins/router/router.module.js
```diff
@@ -130,6 +131,33 @@ internals.registerRoutes = function(server, next) {
...
+        {
+            method: ['GET', 'POST'],
+            path: internals.paths.graphQL,
+            handler: {
+                proxy: {
+                    mapUri: function (req, cb) {
+                        return cb(
+                            null,
+                            `${internals.options.storeUrl}${req.path}`,
+                            Object.assign( // Add 'origin' and 'host' headers to request before proxying
```
bookernath: I originally had this as a nice arrow function with a spread operator, but this made tests fail due to a bug in [eslint/espree](https://github.com/eslint/espree/pull/426) which was only fixed a few months ago. Upgrading our version of espree by 4 major versions seemed prohibitively complicated to make this one change, so I think we should just stick with the weird syntax for now.

bookernath:
Ping @bc-jz @ncheikh

server/plugins/router/router.module.js
ncheikh: Where is `espree` used?

ncheikh: If it is just part of `eslint` you can try bumping it ahead to 6 if that resolves the issue with `espree`. `eslint` doesn't have any functional impact so upgrading it should be very low risk.

bookernath: From my investigation it would mean a serious upgrade of many of the `hapi` libraries, we use, which are pretty fundamental to CLI. I view it as high-risk.

bin/stencil-init.spec.js
```diff
@@ -14,7 +14,7 @@ describe('stencil init', () => {
     let sandbox;
 
     lab.beforeEach(done => {
-        sandbox = Sinon.sandbox.create();
+        sandbox = Sinon.createSandbox();
```
bookernath: Doing this cleanup as part of this PR to fix some deprecation notices. Only affects tests.

PR #550, opened 2020-01-17
Add feature to allow pulling live config files

LordZardeck:
#### What?

Currently if you use `stencil push` to deploy and activate a theme, and you made changes to the theme use the theme editor in the BigCommerce Merchant Admin, you would lose any of those changes. This is problematic, especially when using any kind of CI/CD. This PR attempts to address that by allowing you to easily download the current active theme's config, and overwrite your local one. This can be done in CI/CD before `stencil push`, or manually by a developer if they want to update the repository.


carsonreinke:
@LordZardeck I guess why stop at the `config.json` file, why not the entire theme?

LordZardeck:
@carsonreinke the point of this, is that the only thing (that I am aware of) that changes when between times a developer deploys changes to a theme, is the config file, which changes when a merchant modifies the theme using the theme editor. Everything else would be managed in source control as the authority.

carsonreinke:
@LordZardeck no doubt, but merchant modifications of theme files are definitely possible via the BigCommerce Admin under the Stencil Theme Editor.

bookernath:
@LordZardeck this is great, thanks for putting this in - I'm asking our team to take a look at it next week.

lib/stencil-push.utils.js
```diff
@@ -281,7 +285,8 @@ utils.promptUserWhetherToApplyTheme = (options, callback) => {
         Inquirer.prompt(questions, answers => {
             callback(null, Object.assign({}, options, { applyTheme: answers.applyTheme }));
         });
-    };
+    }
+    ;
```
ncheikh: Can you please collapse the semi-colons and braces back to one line in this file.

ncheikh:
> @LordZardeck no doubt, but merchant modifications of theme files are definitely possible via the BigCommerce Admin under the Stencil Theme Editor.

These changes can definitely happen locally as well, thus leading to multiple changes from 2 different sources. However as we have no diffing functionality at this time, this change will at the very least allow for a more flexible development flow.  Using `pull` will require some care and discipline to ensure one doesn't overwrite their local changes, and I think this is a good call out and something we should note in our documentation.

@carsonreinke thank you for calling this out!

lib/stencil-push.utils.js
LordZardeck: I'll do you one better and remove the semi-colon. My editor tried to auto-fix this styling, but it was confused (as it should be) because this semi-colon isn't supposed to exist. It's trying to terminate the `else` block.

lib/stencil-push.utils.js
```diff
@@ -309,7 +315,8 @@ utils.getVariations = (options, callback) => {
             callback(null, Object.assign({}, options, { variationId: result.variations[0].uuid }));
         } else {
             callback(null, Object.assign({}, options, result));
-        };
+        }
+        ;
```
ncheikh: Please remove this semi-colon as well

lib/stencil-push.utils.js
```diff
@@ -298,7 +303,8 @@ utils.getVariations = (options, callback) => {
     }, (error, result) => {
         if (error) {
             return callback(error);
-        };
+        }
+        ;
```
ncheikh: Please remove this semi-colon as well

lib/stencil-push.utils.js
LordZardeck: done

lib/stencil-push.utils.js
LordZardeck: done

PR #555, opened 2020-02-04
Add schema validation to bundle process

junedkazi:
#### What?
Add schema validation to bundle process.

Screen Shot 2020-02-04 at 1 28 09 PM

@bigcommerce/storefront-team @bigcommerce/merc-team


lib/bundle-validator.js
```diff
@@ -18,7 +18,9 @@ var Eslint = require('eslint').CLIEngine;
 var Fs = require('fs');
 var sizeOf = require('image-size');
 var Path = require('path');
-var Validator = require('jsonschema').Validator;
+var Validator = require('ajv');
+const JsonSchemaValidatorOptions = {schemaId: 'auto', allErrors: true, verbose: true};
```
lord2800: 🍹 Does it make any sense to make any of these configurable? My gut says no, but I figured it's worth asking.

lib/bundle-validator.js
```diff
@@ -66,15 +69,38 @@ BundleValidator.prototype.sizeOf = function (path, callback) {
...
 };
 
+/**
+ * If theme schema exists we need to validate it make sure it passes all defined checks.
+ * @param callback
+ * @returns {*}
+ */
+function validateThemeSchema(callback) {
+    const v = new Validator(JsonSchemaValidatorOptions);
+    var validationSchema = './themeSchema.schema.json';
```
lord2800: `let` here instead of `var`.

lib/bundle-validator.js
junedkazi: We can make it configurable based on feedback in the future.

lib/themeSchema.schema.json
```diff
@@ -0,0 +1,199 @@
+{
```
mattolson: Would prefer to put this in `config/schemas/themeConfig.json` and move `lib/themeConfig.schema.json` as well 🍹

lib/bundle-validator.js
```diff
@@ -66,15 +69,38 @@ BundleValidator.prototype.sizeOf = function (path, callback) {
     return sizeOf(path, callback);
 };
 
+/**
+ * If theme schema exists we need to validate it make sure it passes all defined checks.
+ * @param callback
+ * @returns {*}
+ */
+function validateThemeSchema(callback) {
```
mattolson: Seems like there is a lot of commonality with `validateThemeConfiguration`. Perhaps a refactor is in order?

lib/schemas/privateThemeConfig.json
```diff
@@ -1,5 +1,4 @@
 {
-  "$schema": "http://json-schema.org/draft-04/schema#",
   "id": "http://themes.bigcommerce.com/theme_packages/config",
```
mattolson: Is this schema different from `lib/schemas/themeConfig.json`? If so, it should have a different `id`.

lib/schemas/themeSchema.json
```diff
@@ -0,0 +1,199 @@
+{
+  "$id": "http://themes.bigcommerce.com/theme_packages/schema",
```
mattolson: Maybe we can call this `editorSchema`? `schema` is very generic.

lib/bundle-validator.js
```diff
@@ -86,31 +103,45 @@ function validateThemeConfiguration(callback) {
...
+ * @returns {Error}
+ */
+function validateJsonSchema(type, schema, data) {
+    let errorMessage;
+    const validator = new Validator(JsonSchemaValidatorOptions);
+    validator.validate(schema, data);
+    if (validator.errors && validator.errors.length > 0) {
+        errorMessage = 'Your theme\'s ' + type + '.json has errors:';
+        validator.errors.forEach(error => {
+            errorMessage += '\r\n' + type + error.dataPath + " " + error.message;
```
mattolson: Would be better to use `os.EOL` and anywhere we have `\n` or `\r\n` since it is OS-specific

lib/bundle-validator.js
```diff
@@ -86,31 +103,45 @@ function validateThemeConfiguration(callback) {
...
+ * @param schema
+ * @param data
+ * @returns {Error}
+ */
+function validateJsonSchema(type, schema, data) {
+    let errorMessage;
+    const validator = new Validator(JsonSchemaValidatorOptions);
+    validator.validate(schema, data);
+    if (validator.errors && validator.errors.length > 0) {
+        errorMessage = 'Your theme\'s ' + type + '.json has errors:';
```
mattolson: It's cleaner to use template literals for strings that require interpolation:
```javascript
`Your theme's ${type}.json has errors:`
```

lib/bundle-validator.js
```diff
@@ -86,31 +104,45 @@ function validateThemeConfiguration(callback) {
...
+ * @param schema
+ * @param data
+ * @returns {Error}
+ */
+function validateJsonSchema(type, schema, data) {
+    let errorMessage;
+    const validator = new Validator(JsonSchemaValidatorOptions);
+    validator.validate(schema, data);
+    if (validator.errors && validator.errors.length > 0) {
+        errorMessage = `Your theme\'s ${type}.json has errors:`;
```
mattolson: You don't need to escape the apostrophe here.

mattolson: You can also move the `let errorMessage` into this block

PR #571, opened 2020-02-14
fix: move code to dev dependency

junedkazi:
#### What?
move code to dev dependency since it is required only for specs.

@bigcommerce/storefront-team


junedkazi:
This was just a rebase.

PR #572, opened 2020-02-14
feat: drop node 8 support since it is EOL

junedkazi:
#### What?
drop node 8 support since it is EOL https://github.com/nodejs/Release#end-of-life-releases

@bigcommerce/storefront-team

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,15 +4,14 @@ os:
   - osx
 addons:
   ssh_known_hosts:
-    - github.com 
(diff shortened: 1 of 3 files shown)


carsonreinke:
@junedkazi just Node 10?

junedkazi:
@carsonreinke we are working on adding support for node 12 soon. We still need to work through a few dependencies on our backend infrastructure before we add it on cli.

PR #573, opened 2020-02-14
feat: deprecate theme editor

junedkazi:
#### What?
Since theme editor is replaced with a new version of editor. The current version of theme editor in cli is no longer supported. 

#### Tickets / Documentation
Depends on https://github.com/bigcommerce/stencil-cli/pull/555

@bigcommerce/storefront-team @bigcommerce/merc-team


lib/stencil-bundle.js
```diff
@@ -19,7 +19,6 @@ const PATHS_TO_ZIP = [
     '.scss-lint.yml',
     'stencil.conf.js',
     'templates/**/*',
-    'webpack.*.js',
```
mattolson: We should leave these here so theme downloads are still usable.

server/plugins/renderer/renderer.module.js
```diff
@@ -390,7 +390,7 @@ internals.getPencilResponse = function (data, request, response, configuration)
     data.context.settings.theme_version_id = Utils.int2uuid(1);
     data.context.settings.theme_config_id = Utils.int2uuid(request.app.themeConfig.variationIndex + 1);
     data.context.settings.theme_session_id = null;
-    data.context.settings.maintenance = {secure_path: `http://localhost:${internals.options.stencilEditorPort}`};
+    data.context.settings.maintenance = {secure_path: `http://localhost:8181`};
```
junedkazi: just hardcoding the port here since that was the default if the option for the port was not being passed.

mattolson: What is this used for? And if it's still useful, why not use `stencilServerPort` rather than the `stencilEditorPort` (which is no longer used)?

package.json
```diff
@@ -84,21 +80,11 @@
   "devDependencies": {
     "@commitlint/cli": "^7.2.1",
     "@commitlint/config-conventional": "^7.1.2",
-    "angular": "1.5.6",
```
junedkazi: remove handlbars as well

Issue #376, opened 2018-04-07
Stencil Pull functionality


PeteyRev:
Any plan to perhaps make a `stencil pull` command to get the stores current theme and unzip locally? I work on a lot of different clients' sites in which they will make changes to, so I constantly have to download their current theme and do this manually

Lewiscowles1986:
Yes I've tried all download options and the stencil theme still looks different to the one installed on BigCommerce. It's frustrating to say the least given that I have all permissions.

bookernath:
Great idea! I don't think we have any near-term plans for this, although PRs are welcome.

You can see the docs for our Themes API that we use for `stencil push` here - https://developer.bigcommerce.com/api/v3/themes.html

junedkazi:
Fixed in https://github.com/bigcommerce/stencil-cli/pull/583

PR #583, opened 2020-03-11
feat: added re-integration script for all remote files

martinchuka:
#### What?

This pulls the entire theme from the remote store or just a single file, it overwrites the local theme file(s) if exists or creates a new one, because of the damage this action can cause, I also added a prompt to prevent mistakenly executing this command.

**This is an extension to https://github.com/bigcommerce/stencil-cli/pull/550**

The previous command from https://github.com/bigcommerce/stencil-cli/pull/550 to pull the `config.json` can still be used like `stencil pull --file=config.json`


carsonreinke:
@martinchuka there is a few files that would need to be ignored from the zip file, such as:

- `parsed/`
- `manifest.json`

martinchuka:
> @martinchuka there is a few files that would need to be ignored from the zip file, such as:
> 
> * `parsed/`
> * `manifest.json`

Thanks @carsonreinke, this has been added to exclude list

lib/stencil-pull.utils.js
```diff
@@ -72,31 +75,81 @@ utils.downloadThemeConfig = (options, callback) => {
 
                                     let configFileData = '';
 
+                                    if (options.file && options.file.length) {
+                                        const filePattern = new RegExp((options.file.split(path.sep)).pop()+'$','g');
```
carsonreinke: @martinchuka you want to use `path.parse` ([see](https://nodejs.org/api/path.html#path_path_parse_path)) instead

martinchuka: @carsonreinke - Thank you!. this is exactly what it should have been :+1:

lib/stencil-pull.utils.js
```diff
@@ -72,30 +76,88 @@ utils.downloadThemeConfig = (options, callback) => {
...
+                                        }
+                                        foundMatch = true;
+
+                                    } else if (options.exclude && options.exclude.length) {
+
+                                        /**
+                                         * Do not process any file or directory within the exclude option
+                                         */
+                                        for (let i = 0; i < options.exclude.length; i++) {
+                                            if ((new RegExp(`^${options.exclude[i]}`)).test(entry.fileName)) {
```
carsonreinke: @martinchuka regexp seems like a bad idea here

martinchuka: > @martinchuka regexp seems like a bad idea here

@carsonreinke - I can't use `indexOf` since the directory inside the exclude can be way shorter than the file itself, I understand 'startsWith` and `endsWith` is far better than regex so I used those instead.

carsonreinke:
@martinchuka I think that looks good

carsonreinke:
@junedkazi @bookernath any feedback?

bookernath:
It seems like the intent of `stencil pull` was to quickly sync theme _settings_ which may become out of sync by a merchant's actions in the control panel. We actually want to improve this implementation (for theme settings in particular) with a new API we're building - you can preview it at `/v3/themes/UUID/configurations?site_id:in=1000` (although note it's not considered supported until we've publicly documented it).

The advantage of this is that you'll be able to simply fetch the theme settings from a REST API, without having to perform the download step at all - so it should get this whole process of syncing theme settings from ~30 seconds to under a second.

Should be pretty nice!

With that problem solved, it seems like we can consider this new problem set differently, so I'm curious to get more context on why this is needed.

I'm fine to let it go forward in general, but I think we'll need to use a different command besides `stencil pull` if we want that to still mean "download live settings".

I guess a few options here:

1. Rename the command in this PR

carsonreinke:
@bookernath 

It is pretty common, beyond the theme settings, for template files to be modified directly within the theme editor.  This could even be a problem where an other actor has changed the theme and you simply need to get changed files back into a repo.  Whenever we do a deployment of a theme, we always download first, because our customers always seem to be modifying some element of files.

So, for the name of the command, `push` right now uploads the entire theme, should `pull` not do the opposite?  Either way, I am fine with changing, just as long as we can have this ability to easily do this.

I would say to me, this is a new feature and using `stencil download` would be better instead of introducing a breaking change, which, could always come in the future if BigCommerce wanted. 

cc @martinchuka

martinchuka:
@carsonreinke - Thanks for the review :+1: 

This command is now renamed to `stencil download`

cc @bookernath

bin/stencil-download.js
```diff
@@ -0,0 +1,59 @@
...
+const pkg = require('../package.json');
+const Program = require('commander');
+const stencilDownload = require('../lib/stencil-download');
+const versionCheck = require('../lib/version-check');
+const themeApiClient = require('../lib/theme-api-client');
+const inquirer = require('inquirer');
+
+Program
+    .version(pkg.version)
+    .option('--host [hostname]', 'specify the api host', apiHost)
```
junedkazi, line 16: Can you share some details around how this hostname will be used ?

bin/stencil-download.js
```diff
@@ -0,0 +1,59 @@
...
+    .option('--file [filename]', 'specify the filename to download only')
+    .option('--type [type]', 'specify the type of content to download available are web-pages, theme, web-dav')
```
junedkazi, line 18: I don't see the type being used

bin/stencil-download.js
```diff
@@ -0,0 +1,59 @@
...
+    .parse(process.argv);
+
+if (!versionCheck()) {
+    process.exit(2);
+}
+
+const overwriteType = Program.file ? Program.file : 'files';
+
+Object.assign(options, {
+    exclude: ['parsed', 'manifest.json'],
```
junedkazi, line 28: I think this by default is great but will be a good option as well where I can say give me everything except for a particular folder.

lib/stencil-download.utils.js
```diff
@@ -0,0 +1,137 @@
+const request = require("request");
+const yauzl = require('yauzl');
+const fs = require('fs');
+const tmp = require('tmp');
+const path = require('path');
+const utils = {};
+
+module.exports = utils;
+
+utils.downloadThemeFiles = (options, callback) => {
```
junedkazi, line 10: Would love some tests around this until function

bin/stencil-download.js
martinchuka: @junedkazi - hostname is all used within https://github.com/bigcommerce/stencil-cli/blob/master/lib/theme-api-client.js, and available within most of the command

bin/stencil-download.js
martinchuka: @junedkazi - This shouldn't be part of the option here and it is removed!

bin/stencil-download.js
martinchuka: @junedkazi - 

I don't think there is any reason to have the `parsed` and `manifest` downloaded, so I think maybe we can leave that option and introduce an `exclude` as CLI arg. What do you think?

lib/stencil-download.utils.js
martinchuka: @junedkazi -  You mean the `stencil-download.utils`. Sounds good!. I will put something in for that.

bin/stencil-download.js
junedkazi: Sorry if I was not clear but I was referring to what you are saying 😃

lib/stencil-download.utils.js
```diff
@@ -0,0 +1,137 @@
...
+
+        (
+            new Promise(
+                (resolve, reject) =>
+                    request(options.downloadUrl)
+                        .pipe(fs.createWriteStream(tempThemePath))
+                        .on('finish', () => resolve(tempThemePath))
+                        .on('error', reject),
+            )
+        )
```
junedkazi, line 24: The indentation looks off here on these 2 closing `)`

lib/stencil-download.utils.js
martinchuka: @junedkazi - test has been added!

bin/stencil-download.js
martinchuka: @junedkazi - Thank you!. Added the `exclude` option to CLI

lib/stencil-download.utils.js
martinchuka: @junedkazi - I think the indentation here is okay.

PeteyRev:
#376

martinchuka:
> Awesome @martinchuka #376. I actually scribbled out an [npm package](https://github.com/PeteyRev/stencil-x) to handle this, but obviously this is much more ideal. How did you obtain the store hash? I could not find where it was being stored.

@PeteyRev - Thank you :+1: . The store hash is available within the `options` from this request https://github.com/bigcommerce/stencil-cli/blob/master/lib/stencil-push.utils.js#L52

Issue #585, opened 2020-03-22
Add to cart not working on localhost


flyingL123:
Please help. Originally reported here:

https://github.com/bigcommerce/cornerstone/issues/1653

junedkazi:
@flyingL123 it is due to SameSite https://web.dev/samesite-cookies-explained/ cookies which was introduced in chrome. We are working on a fix but in the interim I would recommend using chrome 76 or less or use any other browser other than chrome.

flyingL123:
Thanks so much for the quick response!

bookernath:
You can also disable SameSite cookie protections in chrome in `chrome://flags` for now.

junedkazi:
@flyingL123 can you please verify if this issue is fixed cli 2.0

cmaycumber:
@junedkazi @flyingL123  I just ran into this issue today and upgrading to v2.0 of the cli fixed it.

junedkazi:
Thank you @cmaycumber for verifying the fix.

PR #586, opened 2020-03-24
fix: address chrome cookie SameSite issue

junedkazi:
#### What?
Fix SameSite cookie issue

@bigcommerce/storefront-team

### DIFF
--- a/server/plugins/renderer/renderer.module.js
+++ b/server/plugins/renderer/renderer.module.js
@@ -246,6 +246,10 @@ internals.parseResponse = function (bcAppData, request, response, responseArgs, 
                     response: response,
                 }, internals.cacheTTL);
 
+                if (response.headers['set-cookie']) {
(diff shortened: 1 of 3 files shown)


junedkazi:
https://github.com/bigcommerce/stencil-cli/issues/585
https://github.com/bigcommerce/cornerstone/issues/1653

server/plugins/renderer/renderer.module.js
```diff
@@ -246,6 +246,10 @@ internals.parseResponse = function (bcAppData, request, response, responseArgs,
                     response: response,
                 }, internals.cacheTTL);
 
+                if (response.headers['set-cookie']) {
```
mattolson: Why do we need this here? We already have it in `getResponse`.

junedkazi: bcoz we cache the info but if it is not found in cache we make the call again. There is a bit of duplication which I plan on fixing in follow up PR's. Just wanted to get this out there for now.

PR #595, opened 2020-04-09
Adding Storefront Plugin and Language

ncheikh:
#### What?

#### Tickets / Documentation

- [Link 1](http://example.com)
- ...

Example Image

### DIFF
--- a/bin/stencil-start
+++ b/bin/stencil-start
@@ -12,6 +12,7 @@ var Pkg = require('../package.json');
 var Program = require('commander');
 var Server = require('../server');
 var ThemeConfig = require('../lib/theme-config');
+var StorefrontConfig = require("@bigcommerce/stencil-cli-storefront-plugin")
 var buildConfig = require('../lib/build-config');
 var Url = require('url');
 var Wreck = require('wreck');
@@ -160,6 +161,7 @@ function startServer() {
             '/assets',
             '/templates',
             '/lang',
+            '/.config'
         ];
         var watchIgnored = [
             '/assets/scss',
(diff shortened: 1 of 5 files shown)


bin/stencil-start
```diff
@@ -228,9 +236,21 @@ function startServer() {
                 ignoreInitial: true,
                 ignored: watchIgnored.map(val => Path.join(themePath, val)),
             },
-            proxy: "localhost:" + stencilServerPort,
+            proxy: {
+                target: "localhost:" + stencilServerPort,
+                proxyReq: [
```
junedkazi, line 241: I am wondering if we can do this inside the plugin itself.

ncheikh: I will test

ncheikh: So we could do this, but I would further argue that we shouldn't.

it looks like this:

```  
bs._options
    .get("proxy")
    .get("proxyReq")
    .push(function (proxyReq: any) {
      proxyReq.setHeader(
        "Accept-Language",
        StorefrontConfig.getInstance().getConfiguredLanguage()
      );
    });
```

the `._options` is intended to be private to browser sync.

Same issue as above in regards to initialization logic all over the place.  We could have collisions and weird overrides.  It will also be harder to track bugs.

junedkazi: Since it is a plugin the logic should be self contained in the plugin itself. The only thing which should come from the client is overrides or settings params.

ncheikh: With further testing this unfortunately does not seem to work if it is in the plugin.  Is this something that can be circled on in future PRs as more logic is added?

PR #599, opened 2020-04-30
feat: watch storefront config file

junedkazi:
#### What?
watch storefront config file and publish an event

@bigcommerce/storefront-team

### DIFF
--- a/bin/stencil-start
+++ b/bin/stencil-start
@@ -160,6 +160,7 @@ function startServer() {
             '/assets',
             '/templates',
             '/lang',
+            '/.config'
         ];
         var watchIgnored = [
             '/assets/scss',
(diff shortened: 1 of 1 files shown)


bin/stencil-start
```diff
@@ -230,7 +243,6 @@ function startServer() {
             },
             proxy: "localhost:" + stencilServerPort,
             tunnel: tunnel,
-            https: Url.parse(staplerUrl).protocol === 'https:'
```
bc-williamkwon, line 233: Is there a reason that we are removing https?

junedkazi: This causes issues with stencil start since it forces the users to use https and http is unavailable. Also it causes issue with browser sync UI to not load

PR #613, opened 2020-08-04
fix: STRF-8574, bump version of "inquirer" package to fix security issues + update .eslintrc

MaxGenash:
#### What?

Bumped version of the outdated npm module "inquirer" to fix security vulnerabilities.
The newer version of the npm module lost the support of callbacks, so I had to update the corresponding parts of our code to use async/await instead of callbacks (the second commit).

Also, updated .eslinrc to allow newer JS syntax, esp. _Object spread operator_ which is supported by NodeJS v10 (the first commit).

#### Tickets / Documentation

This PR is a part of STRF-8574 (since there are going to be a lot of updates they will be split into several PRs).

### DIFF
--- a/lib/stencil-init.js
+++ b/lib/stencil-init.js
@@ -1,18 +1,17 @@
 'use strict';
-var Fs = require('fs');
-var jsonLint = require('./json-lint');
-var Path = require('path');
-var Inquirer = require('inquirer');
-var themePath = process.cwd();
-var hoek = require('hoek');
-var configuration;
-var internals = {};
+const Fs = require('fs');
+const { promisify } = require("util");
+const Path = require('path');
+const Inquirer = require('inquirer');
+const hoek = require('hoek');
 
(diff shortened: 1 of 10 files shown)


lib/stencil-init.js
```diff
@@ -1,18 +1,17 @@
...
+const { promisify } = require("util");
+const Path = require('path');
+const Inquirer = require('inquirer');
+const hoek = require('hoek');
 
-internals.parseAnswers = function(JspmAssembler, ThemeConfig, dotStencilFile, dotStencilFilePath, answers) {
+const jsonLint = require('./json-lint');
+const themePath = process.cwd();
 
+async function performAnswers(JspmAssembler, ThemeConfig, stencilConfig, dotStencilFilePath, answers) {
```
junedkazi, line 11: should this be `parseAnswers`

MaxGenash:
@junedkazi it's ok now, check the PR plz

lib/stencil-init.js
MaxGenash: @junedkazi  I decided to rename the function since it's actually _performing_ already _parsed_ answers :)

PR #616, opened 2020-08-10
fix: strf-8574, remove redundant dependency "hoek"

MaxGenash:
#### What?

Replaced usages of npm module "hoek" with "lodash" to avoid supporting 2 similar libraries.
Did it since "hoek" had security vulnerabilities and required updating the library to newer major versions anyway.

#### Tickets / Documentation

This PR is a part of STRF-8574 (since there are going to be a lot of updates they will be split into several PRs).

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,6 @@
     "good-console": "^4.1.0",
     "graceful-fs": "^4.1.11",
     "hapi": "^8.4.0",
-    "hoek": "^2.12.0",
     "image-size": "^0.4.0",
     "inquirer": "^7.3.3",
     "jsonlint": "^1.6.2",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -54,7 +54,6 @@
     "good-console": "^4.1.0",
     "graceful-fs": "^4.1.11",
     "hapi": "^8.4.0",
-    "hoek": "^2.12.0",
```
junedkazi, line 57: No package lock changes after dropping this package ?

MaxGenash: @junedkazi yes, I double-checked.
Probably that’s because Hapi and its modules use "hoek" internally.

PR #617, opened 2020-08-13
fix: strf-8574, bump "hapi" and its modules to fix security issues

MaxGenash:
#### What?

Bumped outdated versions of "hapi" and its related modules to fix security vulnerabilities.

Modules "h2o2" and "inert" were emitted from "hapi" in the newer versions, so I added them as separate dependencies.

#### Tickets / Documentation

This PR is a part of STRF-8574


bin/stencil-start
```diff
@@ -1,34 +1,26 @@
...
+const ThemeConfig = require('../lib/theme-config');
+const buildConfig = require('../lib/build-config');
+const Url = require('url');
+const Wreck = require('wreck');
+const jsonLint = require('../lib/json-lint');
+const versionCheck = require('../lib/version-check');
+let themePath = process.cwd();
+let templatePath = Path.join(themePath, 'templates');
+let dotStencilFilePath = Path.join(themePath, '.stencil');
+let themeConfigPath = Path.join(themePath, 'config.json');
```
junedkazi, line 23: Is there a reason why these paths can't be const ?

bin/stencil-start
```diff
@@ -147,116 +135,111 @@ Wreck.get(
...
-            port: browserSyncPort,
-            files: watchFiles.map(val => Path.join(themePath, val)),
-            watchOptions: {
-                ignoreInitial: true,
-                ignored: watchIgnored.map(val => Path.join(themePath, val)),
-            },
-            proxy: "localhost:" + stencilServerPort,
-            tunnel: tunnel,
-        });
+    const tunnel = typeof Program.tunnel === 'undefined'
```
junedkazi, line 214: we can move this to the top where we define other constants 🍹

package.json
```diff
@@ -36,11 +36,17 @@
     "@bigcommerce/stencil-paper": "^3.0.0-rc.29",
     "@bigcommerce/stencil-styles": "1.2.1",
     "@octokit/rest": "^18.0.3",
+    "@hapi/boom": "^8.0.1",
+    "@hapi/glue": "^6.2.0",
+    "@hapi/good": "^8.2.4",
+    "@hapi/good-console": "^8.1.2",
```
junedkazi, line 42: This is going to be deprecated soon. See https://www.npmjs.com/package/@hapi/good-console for details.
This is not a blocker but we should investigate and figure out if it is required or we drop the usage in a followup PR.

package.json
```diff
@@ -36,11 +36,17 @@
     "@bigcommerce/stencil-paper": "^3.0.0-rc.29",
     "@bigcommerce/stencil-styles": "1.2.1",
     "@octokit/rest": "^18.0.3",
+    "@hapi/boom": "^8.0.1",
+    "@hapi/glue": "^6.2.0",
+    "@hapi/good": "^8.2.4",
```
junedkazi, line 41: This is going to be deprecated soon. See https://www.npmjs.com/package/@hapi/good for details.
This is not a blocker but we should investigate and figure out if it is required or we drop the usage in a followup PR.

server/plugins/renderer/responses/raw-response.spec.js
```diff
@@ -19,11 +19,12 @@ lab.describe('RawResponse', () => {
     const statusCode = 200;
     let request;
     let response;
-    let reply;
+    let h;
```
junedkazi, line 22: what does `h` stand for ?

server/plugins/renderer/responses/raw-response.js
```diff
@@ -14,24 +14,24 @@ const internals = {
  */
 function RawResponse(data, headers, statusCode) {
 
-    this.respond = function (request, reply) {
-        var payload = data;
+    this.respond = function (request, h) {
```
junedkazi, line 17: what does h stand for ?

bin/stencil-start
MaxGenash: no, I just didn’t notice that they aren’t further modified

bin/stencil-start
MaxGenash: yeah

server/plugins/renderer/responses/raw-response.js
MaxGenash: @junedkazi [From the docs](https://hapi.dev/api?v=20.0.0#response-toolkit) "h" stand for HAPI.
They replaced the _reply_ function with this "Response Toolkit" (h) in v17.

server/plugins/renderer/responses/raw-response.spec.js
MaxGenash: @junedkazi From the docs "h" stand for HAPI.
They replaced the reply function with this "Response Toolkit" (h) in v17.

package.json
MaxGenash: Ok, I'll check it

package.json
MaxGenash: Ok, I'll check it

bin/stencil-start
MaxGenash: updated

bin/stencil-start
MaxGenash: updated

package.json
MaxGenash: @junedkazi I checked, we can drop "@hapi/good" and "@hapi/good-console" since they aren't actually used in this project. I can create a corresponding followup PR when this one gets merged.

PR #620, opened 2020-08-18
STRF-8589 drop jspm support

MaxGenash:
#### What?

Dropped JSPM support.

#### Tickets / Documentation

https://jira.bigcommerce.com/browse/STRF-8589


bin/stencil-start
```diff
@@ -29,7 +28,6 @@ Program
     .version(Pkg.version)
     .option('-o, --open', 'Automatically open default browser')
     .option('-v, --variation [name]', 'Set which theme variation to use while developing')
-    .option('-t, --test', 'Enable QA mode which will bundle all javascript for speed to test locally')
```
MaxGenash, line 32: @junedkazi note, that after removing code related to jspm, this CLI option has no effect, so I removed it too.
We will need to update the [docs](https://developer.bigcommerce.com/stencil-docs/installing-stencil-cli/stencil-cli-options-and-commands#stencil-start) to remove this option from there as well.

junedkazi: @bigcommerce/dev-docs

PR #627, opened 2020-08-28
feature(storefront): STRF-8629 Update cli to use template engine based on the value stored in config.json

jairo-bc:
…config.json

#### What?

#### Tickets / Documentation

- [Link 1](http://example.com)
- ...

Example Image

### DIFF
--- a/server/plugins/renderer/renderer.module.js
+++ b/server/plugins/renderer/renderer.module.js
@@ -359,6 +359,7 @@ internals.getTemplatePath = function (path, data) {
  */
 internals.getPencilResponse = function (data, request, response, configuration) {
     data.context.theme_settings = configuration.settings;
+    data.context.template_engine = configuration.template_engine;
 
     // change cdn settings to serve local assets
     data.context.settings.cdn_url = '';
--- a/server/plugins/renderer/responses/pencil-response.js
+++ b/server/plugins/renderer/responses/pencil-response.js
@@ -1,48 +1,36 @@
 const _ = require('lodash');
 const Paper = require('@bigcommerce/stencil-paper');
-const internals = {};
 
-module.exports = function (data, assembler) {
-    this.respond = function (request, h) {
-        const paper = new Paper(data.context.settings, data.context.theme_settings, assembler, "handlebars-v3");
-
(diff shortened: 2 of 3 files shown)


server/plugins/renderer/renderer.module.js
```diff
@@ -359,6 +359,7 @@ internals.getTemplatePath = function (path, data) {
  */
 internals.getPencilResponse = function (data, request, response, configuration) {
     data.context.theme_settings = configuration.settings;
+    data.contenxt.template_engine = configuration.settings.template_engine || "handlebars-v3";
```
mattolson, line 362: `contenxt` is a typo. Also, I don't think we're going to want to put this in the "settings" object. I think we should put it into a top-level key in https://github.com/bigcommerce/cornerstone/blob/master/config.json.

server/plugins/renderer/responses/pencil-response.spec.js
```diff
@@ -0,0 +1,67 @@
...
+            code: () => response,
+        };
+
+        h = {
+            response: sinon.stub().returns(response),
+        };
+
+    });
+
+    it('should return error, when the wrong template_engine is sent', () => {
```
junedkazi, line 53: We need to test for 3 cases;

- When the key doesn't exist `->` In this case it should default to `handlebars-v3`
- When the key exists but it doesn't match a predefined list `->` It should throw an error
- When key exists and it matches predefined list `->` It should render using the passed template engine

server/plugins/renderer/responses/pencil-response.js
```diff
@@ -65,56 +53,82 @@ internals.getTemplatePath = function (request, data) {
...
 
-/**
- * Scape special characters for regular expression
- *
- * @param string
- */
-internals.escapeRegex = function (string) {
-    return string.replace(/[-\\^$*+?.()|[\]{}]/g, "\\$&");
+        if (!isSupportedHandlebarsVersion(data.context.template_engine)) {
+            throw new Error('Provided Handlebars version is not supported!');
```
junedkazi, line 99: We should also try and list the supported versions

PR #634, opened 2020-09-11
feat: make init command more suitable for automation

NickTolhurst26:
#### What?

To make this cli tool more 'automatable' it's necessary to remove the dependency of user input when using the init command.

Rather than having a flag that bypasses user input, it would seem more appropriate to have the cli not ask for values that are already provided via the cli options.

This pull request removes the cli options from the 'defaultAnswers' and uses them to determine which prompts are necessary or not.

It then merges the prompt answers with the cli answers to return the same object as before.

Please let me know if I have done anything incorrect in this contribution process as this is the first time I've contributed to a project - feedback is always welcome.


jairo-bc:
I think it's better to add a parameter to bypass prompts. 
Another thing is, that for some reason eslint is not validating. Do you have any ideas why? @MaxGenash

jairo-bc (commented):
I think it's better to add a parameter to bypass prompts.
Another thing is, that for some reason eslint is not validating. Do you have any ideas why? @MaxGenash

lib/stencil-init.js
```diff
@@ -115,8 +133,13 @@ class StencilInit {
                         return true;
                     }
                 },
-            },
-        ]);
+            });
+        }
+
+        return { 
+            ...await this.inquirer.prompt(prompts),
```
MaxGenash, line 140: We are calling `inquirer.prompt` anyway even if all the answers are provided as CLI options, but the idea of the PR was (if I understand it correctly) to avoid calling inquirer.prompt.

MaxGenash:
@NickTolhurst26 thanks for the interest in contributing to this project.

From the description of the PR it's said that we are going to make the CLI command more suitable for automation by avoiding user input and using CLI options instead. Are you going to use some testing tools to run CLI commands directly to have real e2e tests? Because at the moment I don’t see any substantial improvements in tests - we are still mocking everything with jest.

MaxGenash:
Just to note: if we accept this PR, I think we should apply similar changes (don’t ask users to input values that were already provided as CLI options) to all the CLI commands to avoid inconsistency in bahavior

lib/stencil-init.js
NickTolhurst26: Yeah, you're right. Let me take another look. :)

NickTolhurst26:
@MaxGenash - I hadn't thought about how to perform e2e tests of a node CLI. Do you know of any particular tools that could achieve this? I've had a browse around and haven't found any strong candidates. I've found a tool called [nixt](https://www.npmjs.com/package/nixt). Is this sort-of what you meant?

My use-case for this pull request is that I have a CD pipeline deploying to BigCommerce using `stencil-cli` in Azure DevOps. Most of it is done with stencil-cli, but when it comes to using `stencil init` it fails on waiting for user input so I'm having to generate the `.stencil` JSON file in PowerShell instead. It's a step better than checking in the `.stencil` file, but ultimately It'd be nice to see an stencil-cli task in the AzDo marketplace. 

Also, I agree that it makes sense to apply these changes across the board. I'll see what I can do. I'm fairly new when it comes to javascript and the `stencil init` command seems fairly lightweight compared to the rest.

NickTolhurst26:
> Just to note: if we accept this PR, I think we should apply similar changes (don’t ask users to input values that were already provided as CLI options) to all the CLI commands to avoid inconsistency in bahavior

@MaxGenash - looks like only `stencil push` and `stencil init` call `inquirer.prompt`. Stencil push seems to be handed with the `--activate` and `--delete` flags.

MaxGenash:
> My use-case for this pull request is that I have a CD pipeline deploying to BigCommerce using `stencil-cli` in Azure DevOps. Most of it is done with stencil-cli, but when it comes to using `stencil init` it fails on waiting for user input so I'm having to generate the `.stencil` JSON file in PowerShell instead.

Oh, ok, I got it. I thought on "more 'automatable'" you meant improving testing tools :)
Your use case sounds reasonable as for me

MaxGenash:
> > Just to note: if we accept this PR, I think we should apply similar changes (don’t ask users to input values that were already provided as CLI options) to all the CLI commands to avoid inconsistency in bahavior
> 
> @MaxGenash - looks like only `stencil push` and `stencil init` call `inquirer.prompt`. Stencil push seems to be handed with the `--activate` and `--delete` flags.

Yeah, indeed. Moreover, in other CLI commands we already don’t ask users to input values that were already provided as CLI options (e.g. [stencil-push --delete](https://github.com/bigcommerce/stencil-cli/blob/master/lib/stencil-push.utils.js#L160),  [stencil-push --activate](https://github.com/bigcommerce/stencil-cli/blob/master/lib/stencil-push.utils.js#L272) ), so stencil-init seems to be the only place to update

lib/stencil-init.js
```diff
@@ -70,38 +85,46 @@ class StencilInit {
...
+            accessToken: stencilConfig.accessToken,
+            port: stencilConfig.port || this.serverConfig.get('/server/port'),
         };
     }
 
     /**
      * @param {{port: (number), normalStoreUrl: (string), accessToken: (string)}} defaultAnswers
-     * @returns {Promise<object>}
+     * @param {{port: (number), normalStoreUrl: (string), accessToken: (string)}} cliConfig
+     * @returns {{port: (number), normalStoreUrl: (string), accessToken: (string)}}
```
MaxGenash, line 101: Seems like we no longer return the object, but `object[] | null` instead.

MaxGenash: Actually I would return an empty array rather than null in case of no questions for the simplicity of the types (and then use `questions.lenght ? await this.askQuestions(questions) : {}`), but it is optional since not a mistake, but just a personal preference.

lib/stencil-init.js
```diff
@@ -115,19 +138,31 @@ class StencilInit {
...
+        return prompts.length > 0 ? prompts : null;
+    }
+
+    /**
+     * @param {{port: (number), normalStoreUrl: (string), accessToken: (string)}} defaultAnswers
+     * @param {{port: (number), url: (string), token: (string)}} cliOptions
+     * @returns {Promise<object>}
+     */
+    async askQuestions (questions) {
+        return await this.inquirer.prompt(questions);
```
MaxGenash, line 153: As for me, now this method looks a bit redundant so we can use `this.inquirer.prompt` directly.

lib/stencil-init.js
```diff
@@ -37,15 +37,30 @@ class StencilInit {
...
 
         this.logger.log('You are now ready to go! To start developing, run $ ' + 'stencil start'.cyan);
     }
 
+    /**
+     * @param {{port: (number), url: (string), token: (string)}} cliOptions
+     * @returns {{port: (number), normalStoreUrl: (string), accessToken: (string)}}
+     */
+    getCliConfig (cliOptions) {
+        return {
```
MaxGenash, line 57: Let's better accept cliOptions in the needed format from the very beginning, so we don’t need to update their shape here. Kinda:

*/bin/stencil-init.js:*
```
...
new StencilInit().run(dotStencilFilePath, {
    normalStoreUrl: cliOptions.url,
    accessToken: cliOptions.token,
    port: cliOptions.port,
});
...
```

lib/stencil-init.js
```diff
@@ -115,19 +138,31 @@ class StencilInit {
...
+        return await this.inquirer.prompt(questions);
     }
 
     /**
      * @param {object} stencilConfig
      * @param {object} answers
      * @returns {object}
      */
-    applyAnswers (stencilConfig, answers) {
+    applyAnswers (stencilConfig, answers, cliOptions) {
```
MaxGenash, line 161: Parameter cliOptions is not described in JSDoc

lib/stencil-init.js
```diff
@@ -37,15 +37,30 @@ class StencilInit {
...
     async run (dotStencilFilePath, cliOptions = {}) {
+
+        const cliConfig = this.getCliConfig(cliOptions);
         const oldStencilConfig = this.readStencilConfig(dotStencilFilePath);
-        const defaultAnswers = this.getDefaultAnswers(oldStencilConfig, cliOptions);
-        const answers = await this.askQuestions(defaultAnswers);
-        const updatedStencilConfig = this.applyAnswers(oldStencilConfig, answers);
+        const defaultAnswers = this.getDefaultAnswers(oldStencilConfig);
+        const questions = this.getQuestions(defaultAnswers, cliConfig);
+        const answers = questions ? await this.askQuestions(questions) : {};
```
MaxGenash, line 45: Let's move this ternary operator inside `askQuestions`. It will be easier to test it there. And the `run` method will just contain the integration code with no logic

MaxGenash: kinda:
```
async askQuestions (questions) {
    return questions.length ? this.inquirer.prompt(questions) : {};
}
```

lib/stencil-init.js
```diff
@@ -115,19 +138,31 @@ class StencilInit {
...
+            });
+        }
+
+        return prompts.length > 0 ? prompts : null;
+    }
+
+    /**
+     * @param {{port: (number), normalStoreUrl: (string), accessToken: (string)}} defaultAnswers
+     * @param {{port: (number), url: (string), token: (string)}} cliOptions
+     * @returns {Promise<object>}
```
MaxGenash, line 150: JSDoc params are outdated

MaxGenash:
@NickTolhurst26 My apologies for the long response, had some other urgent stuff to complete.

Generally, the PR LGTM. Left just a few minor remarks and after they are fixed I think it's good to go

MaxGenash (approved):

MaxGenash:
@junedkazi reviewed, looks ok for me. If it's ok for you too - let's merge it

jairo-bc (approved):

jairo-bc (approved):

NickTolhurst26:
Thanks for your time on this guys! And thanks for your advice throughout, @MaxGenash ! 

What is the release process for this? I'm keen to start using the new feature.

Do I just follow the instructions from the `CONTRIBUTING.md` (below)?
> After the changes are merged to master, pull the latest to your local environment, run gulp release and follow the prompts.

Or is this something one of you guys do?

junedkazi:
@NickTolhurst26 we are planning on releasing it early next week. We have a couple more PR we want to see through before we cut a release.

PR #636, opened 2020-09-14
fix: (strf-5280) Multiple themes - Changes to config.json not reflected

jairo-bc:
#### What?
On config.json current variant is being reset.

#### Tickets / Documentation

https://jira.bigcommerce.com/browse/STRF-5280

### DIFF
--- a/bin/stencil-start.js
+++ b/bin/stencil-start.js
@@ -160,6 +160,13 @@ async function startServer() {
         }
     });
 
+    Bs.watch('config.json', event => {
+        if (event === 'change') {
+            themeConfig.resetVariationSettings();
+            Bs.reload();
+        }
+    });
+
     Bs.watch('.config/storefront.json', (event, file) => {
         if (event === 'change') {
             console.log("storefront json changed");
--- a/lib/theme-config.js
+++ b/lib/theme-config.js
@@ -122,6 +122,15 @@ ThemeConfig.prototype.setThemePath = function (themePath) {
 };
 
 /**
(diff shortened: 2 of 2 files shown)


bin/stencil-start.js
```diff
@@ -160,6 +160,18 @@ async function startServer() {
         }
     });
 
+    Bs.watch('config.json', (event) => {
+        if (event === 'change') {
+            try {
+                themeConfig.resetVariationSettings();
```
MaxGenash, line 166: Just curious when `themeConfig.resetVariationSettings()` can throw errors?

jairo-bc: haha, never :) That was left from previous trials using `updateConfig()`

MaxGenash (approved):

lib/theme-config.js
```diff
@@ -121,6 +121,15 @@ ThemeConfig.prototype.setThemePath = function (themePath) {
...
 
+/**
+ * Reset Current Variation Settings
+ *
+ * @returns {ThemeConfig}
+ */
+ThemeConfig.prototype.resetVariationSettings = function () {
+    this.currentVariationSettings = null;
+};
+
```
junedkazi, line 132: do we need to return something here ?

jairo-bc: Nope. It's just a setter. Some methods here return `this` for chaining, but I haven't found any occurrences of this.

junedkazi (approved):

PR #641, opened 2020-09-21
feat: (strf-8674) update "async" npm package

MaxGenash:
#### What?

Updated "async" npm package. Figured out that there were almost no breaking changes for us, [apart form one test](https://github.com/bigcommerce/stencil-cli/compare/master...MaxGenash:STRF-8674_update_async?expand=1#diff-1e2ce63b00bff55f2e0bb165bd286bc5R45). While reviewing all the usages of the library I removed redundant usages of async.nextTick. Also, since in async.js v3 all methods return promise if no callback was passed - I removed promisify in ` bin/stencil-download.js`.

Actually, at the moment maojirty (if not all) of the usages of async.js can be replaced with native means (async/await, Promise.all, for...await..of, etc.), but it will take several days, so I didn’t do it here, but we can do it in a separate ticket.

#### Tickets / Documentation
 
STRF-8674

### DIFF
--- a/bin/stencil-download.js
+++ b/bin/stencil-download.js
@@ -3,7 +3,6 @@
 require('colors');
 const inquirer = require('inquirer');
 const program = require('../lib/commander');
-const { promisify } = require("util");
 
 const { API_HOST, PACKAGE_INFO, DOT_STENCIL_FILE_PATH } = require('../constants');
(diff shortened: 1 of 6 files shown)


jairo-bc (approved):

junedkazi (approved):
@MaxGenash this needs a rebase

PR #645, opened 2020-09-22
feat: (strf-8625) support Node 12.x

MaxGenash:
#### What?

Added support for node 12.x.

Smoke-tested all the CLI commands + ran local store, seems to work fine.

#### Tickets / Documentation

strf-8625


.travis.yml
```diff
@@ -6,7 +6,7 @@ addons:
   ssh_known_hosts:
     - github.com
 node_js:
-  - 10
```
junedkazi, line 9: I wonder if we should support 10 as well until EOL for node 10 or should we switch over support to 12 only
@bookernath thoughts ?

.travis.yml
```diff
@@ -6,7 +6,7 @@ addons:
...
+  - 12.18.4
```
junedkazi, line 9: We also need to update appveyor https://github.com/bigcommerce/stencil-cli/blob/e0b638709a8e51fb420fc8cca17357381144f01f/appveyor.yml#L3

junedkazi (commented):
I would also recommend a clean install to confirm things are working correctly specially making sure `node-sass` is cleanly installed.

MaxGenash:
@junedkazi we should also update the [documentation](https://developer.bigcommerce.com/stencil-docs/installing-stencil-cli/installing-stencil) to reflect the latest version.

Actually instead of
```
nvm install 10.16
nvm use 10.16
```
we can write
```
nvm install
nvm use
```
it will grab the correct version from .nvmrc automatically

.travis.yml
MaxGenash: I did it, it should be visible in https://github.com/bigcommerce/stencil-cli/pull/645/files

MaxGenash:
> I would also recommend a clean install to confirm things are working correctly specially making sure `node-sass` is cleanly installed.

Yeah, I did it

.travis.yml
MaxGenash: We should take into account that there are some dependencies (esp. related to hapi) that have last versions already not supporting node 10

junedkazi (approved):

jairo-bc (approved):

.travis.yml
MaxGenash: Decided that we should support node 10 too

junedkazi: @MaxGenash lets keep node 10 for now since it is an LTS version. We will drop it next year.

MaxGenash: Ok.
@junedkazi also just for curiosity tried node 14, but unfortunately, our old version of node-sass doesn’t work with it :(

junedkazi: yes since we are on a forked version of node-sass we have not compiled a node 14 binary for it.

README.md
```diff
@@ -7,7 +7,9 @@
 The BigCommerce server emulator for local theme development.
 
 ## Install
-_Note: Stencil requires the Node.js runtime environment, version 10.x We do not yet have support for Node 12 or greater._
+_Note: Stencil requires the **Node.js** runtime environment, version **12.x** (recommended),
+or 10.x (still supported, but will be removed soon).
+We do not ensure correct work on other versions at the moment._
```
junedkazi, line 12: I don't think we need this line.

README.md
```diff
@@ -7,7 +7,9 @@
 The BigCommerce server emulator for local theme development.
 
 ## Install
-_Note: Stencil requires the Node.js runtime environment, version 10.x We do not yet have support for Node 12 or greater._
+_Note: Stencil requires the **Node.js** runtime environment, version **12.x** (recommended),
+or 10.x (still supported, but will be removed soon).
```
junedkazi, line 11: How about we change this to
```
_Note: Stencil requires the Node.js runtime environment, version 10.x and 12.x (Recommended) are supported. We do not yet have support for versions greater than Node 12._
```

.travis.yml
MaxGenash: ok, got it

README.md
MaxGenash: ok, updated

junedkazi (approved):

PR #646, opened 2020-09-26
feat: update "stencil pull" to use configurations API, improving performance

bookernath:
#### What?

Building upon @LordZardeck's work in https://github.com/bigcommerce/stencil-cli/pull/550.

Now `stencil pull` will use the new Theme Configurations API to pull the config, instead of downloading the entire theme and plucking out the file.

Improvements:

- MUCH faster (takes only a second or two)
- Supports multiple storefronts via channel ID
- Supports plucking either the "live" configuration or the "saved" (draft) configuration saved in Page Builder
- Only overwrites `settings` keys in config.json, not the whole file
  - If the old behavior is preferred, you can use `stencil download -f config.json`

API changes:

- `--save` argument renamed to `--filename` to avoid confusion with `--saved` (**breaking change**, will go out with 3.0 release)
- New `--channel_id` argument to specify which channel to fetch configuration from. Defaults to 1.
- New `--saved` argument specifies that you want the saved configuration instead of the live one (this is created if you click "save" and not "publish" in Page Builder)
- Shorthands added
  - `-c` for `--channel_id`
  - `-s` for `--saved`
  - `-h` for `--host`
  - `-f` for `--filename`

image

ping @bigcommerce/storefront-team @bigcommerce/merc-team @junedkazi

### DIFF
--- a/bin/stencil-pull.js
+++ b/bin/stencil-pull.js
@@ -6,12 +6,14 @@ const { DOT_STENCIL_FILE_PATH, PACKAGE_INFO, API_HOST } = require('../constants'
 const program = require('../lib/commander');
 const stencilPull = require('../lib/stencil-pull');
 const versionCheck = require('../lib/version-check');
-const { printCliResultError}  = require('../lib/cliCommon');
+const { printCliResultErrorAndExit }  = require('../lib/cliCommon');
 
 program
     .version(PACKAGE_INFO.version)
-    .option('--host [hostname]', 'specify the api host', API_HOST)
-    .option('--save [filename]', 'specify the filename to save the config as', 'config.json')
+    .option('-s, --saved', 'get the saved configuration instead of the active one')
+    .option('-h, --host [hostname]', 'specify the api host', API_HOST)
+    .option('-f, --filename [filename]', 'specify the filename to save the config as', 'config.json')
+    .option('-c, --channel_id [channelId]', 'specify the channel ID of the storefront to pull configuration from', parseInt)
     .parse(process.argv);
 
 if (!versionCheck()) {
@@ -22,13 +24,13 @@ const cliOptions = program.opts();
 const options = {
     dotStencilFilePath: DOT_STENCIL_FILE_PATH,
     apiHost: cliOptions.host || API_HOST,
-    saveConfigName: cliOptions.save,
+    saveConfigName: cliOptions.filename,
+    channelId: cliOptions.channel_id || 1,
+    saved: cliOptions.saved || false,
 };
 
-stencilPull(options, (err, result) => {
+stencilPull(options, err => {
     if (err) {
-        printCliResultError(err);
-        return;
+        printCliResultErrorAndExit(err);
     }
-    console.log('ok'.green + ` -- Pulled active theme config to ${result.saveConfigName}`);
 });
--- a/lib/cliCommon.js
+++ b/lib/cliCommon.js
@@ -24,7 +24,18 @@ function printCliResultError (error) {
     console.log(messages.submitGithubIssue);
 }
 
+/**
+ * @param {Error} error
+ * @returns {void}
+ */
+function printCliResultErrorAndExit (error) {
+    printCliResultError(error);
+    // Exit with error code so automated systems recognize it as a failure
+    process.exit(1);
+}
+
 module.exports = {
     printCliResultError,
+    printCliResultErrorAndExit,
     messages,
 };
--- a/lib/stencil-pull.utils.js
+++ b/lib/stencil-pull.utils.js
@@ -1,54 +1,79 @@
 const themeApiClient = require('./theme-api-client');
-const tmp = require('tmp-promise');
-const { extractZipFiles } = require('./archiveManager');
-const { fetchFile } = require('./utils/networkUtils');
+const fs = require('fs');
+const _ = require('lodash');
 
 const utils = {};
 
 module.exports = utils;
 
-utils.selectActiveTheme = (options, callback) => {
-    const [activeTheme] = options.themes.filter(theme => theme.is_active).map(theme => theme.uuid);
+utils.getChannelActiveTheme = async options => {
+    const { config: { accessToken }, apiHost, storeHash, channelId } = options;
 
-    callback(null, Object.assign({}, options, { activeTheme }));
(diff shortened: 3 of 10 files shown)


lib/stencil-pull.utils.js
```diff
@@ -1,54 +1,79 @@
...
-    }
+    let rawConfig = fs.readFileSync('config.json');
+    let parsedConfig = JSON.parse(rawConfig);
+    let diffDetected = false;
 
-    console.log('ok'.green + ' -- Theme files downloaded');
-    console.log('ok'.green + ' -- Extracting theme config');
+    // For any keys the remote configuration has in common with the local configuration,
+    // overwrite the local configuration if the remote configuration differs
+    for (const [key, value] of Object.entries(remoteThemeConfiguration.settings)) {
```
MaxGenash, line 44: @bookernath why do we take into account only the `config.settings` field?
Can't other fields be changed? (e.g. `config.variations.settings`).

lib/stencil-pull.utils.js
```diff
@@ -1,54 +1,79 @@
...
+        if (key in parsedConfig.settings) {
+            // Check for different types, and throw an error if they are found
+            if (typeof parsedConfig.settings[key] !== typeof value) {
```
MaxGenash, line 47: What the user should do if the types were changed (e.g. `string[]` to `string` or `string` to `number`) and the user can't update the local config?

lib/stencil-pull.utils.js
MaxGenash: Can we just overwrite the file without checking the difference (it will be faster and simpler)?

lib/stencil-pull.utils.js
```diff
@@ -1,54 +1,79 @@
...
+    let rawConfig = fs.readFileSync('config.json');
+    let parsedConfig = JSON.parse(rawConfig);
+    let diffDetected = false;
 
-    console.log('ok'.green + ' -- Theme files downloaded');
-    console.log('ok'.green + ' -- Extracting theme config');
+    // For any keys the remote configuration has in common with the local configuration,
+    // overwrite the local configuration if the remote configuration differs
+    for (const [key, value] of Object.entries(remoteThemeConfiguration.settings)) {
+        if (key in parsedConfig.settings) {
```
MaxGenash, line 45: Seems like we take into account only the fields that our local config had. Can't some new fields be added to the remote config?

bin/stencil-pull.js
```diff
@@ -22,13 +24,14 @@ const cliOptions = program.opts();
...
+    channelId: cliOptions.channel_id || 1,
+    saved: cliOptions.saved || false,
 };
 
-stencilPull(options, (err, result) => {
+stencilPull(options, err => {
     if (err) {
         printCliResultError(err);
-        return;
+        throw err;
```
MaxGenash, line 35: Throwing the error will print the same message as printCliResultError already did (just without the call stack)

lib/stencil-pull.utils.js
bookernath: @MaxGenash thanks for your questions/comments.

The config in the remote may actually have different keys, so we can't just save it directly.

For example, we may have this case...

Remote configuration
```
{
  "settings": {
      "a": 1,
      "b":  8
  }
}
```

Local configuration
```
{
  "settings": {
      "b":  2,
      "c": 3
  }
}
```

In which case the correct "merged" configuration will prefer the remote configuration for any keys in common, ignoring keys that don't exist locally. So we'd want:

Merged configuration
```
{
  "settings": {
      "b":  8,
      "c": 3
  }
}
```

Couple of assumptions

1. If a key is missing from the local configuration, it's not used in the local theme, so it's not needed
2. If a key is in both configurations but has a different value, we want to update the local theme with the remote configuration so that `stencil push` doesn't overwrite the remote configuration (this is the point of `stencil pull`)
3. If a key is missing from the remote configuration that is present in the local configuration, it's probably a new key that was added - so the local theme needs it, but the remote theme doesn't
4. If we have a key in common with a different data type, this is a breaking change and the user needs to know about it via an error

lib/stencil-pull.utils.js
bookernath: Yes, this probably means a mistake was made - so I want to throw an error here to let the user know for manual intervention.

lib/stencil-pull.utils.js
bookernath: New fields are added to the remote config by uploading a new theme that has additional fields. We assume the user is working on that theme locally with CLI - so if new fields are going to be added, they probably already exist in the local theme but not the remote.

Essential the local theme is the source-of-truth for future keys, but not source of truth for their values (as changes may have been made in CP - `stencil pull` helps you get them in sync :)

bin/stencil-pull.js
bookernath: Good idea - I'll make an update.

lib/stencil-pull.utils.js
bookernath: Regarding *variations*, these just act as presets for settings which ultimately end up in config.json. But if you make a change in Page Builder which saves an explicit setting for one of these keys, that value takes precedence. So correct behavior for `stencil pull` is to make `config.json` match the remote configuration (ignoring variations) to make sure storefront behavior is not changed by the next deployment via `stencil pull`

bin/stencil-pull.js
junedkazi: @MaxGenash @bookernath I think we should update `printCliResultError` to also include the stack if it is available and after logging instead of doing a return we can do exit.

bookernath: @MaxGenash @junedkazi ♻️

bookernath: I opted to not do a  full stack trace, because for these "expected" errors it's not super helpful. It just shows a trace back to the async module.

lib/cliCommon.js
```diff
@@ -24,7 +24,19 @@ function printCliResultError (error) {
     console.log(messages.submitGithubIssue);
 }
 
+/**
+ * @param {Error} error
+ * @param {Array<{message: string}>} [error.messages]
```
junedkazi, line 29: there is a single input param. Why are we listing `error.message` seperately here ?

bookernath: Lazy copy paste :P

junedkazi (approved):

PR #652, opened 2020-10-06
fix: (strf-8705) fix broken headers and cookies in local server

MaxGenash:
#### What?

Fixed bugs in the local server:
1. Local server always returned empty headers because of the typo [in the code](https://github.com/bigcommerce/stencil-cli/compare/master...MaxGenash:STRF-8705_Bug_local_server_is_broken_after_updating_packages?expand=1#diff-596ce37f7ed6dc6ea50288ac9903be41L393) (should be `_.fromPairs(...response.headers.entries())` instead of `_.fromPairs(response.headers.entries())`).
2. Response headers 'set-cookies' were joined into a single string because of [a problem in node-fetch implementation](https://github.com/node-fetch/node-fetch/issues/251) which resulted in parsing the cookies by the browser incorrectly. To avoid it I started using response.headers.raw() which returns an internal raw map of headers.
3. Figured out that fetch handles redirects automatically by default in contrast with the previous library (wreck), which resulted in skipping processing of the redirect response headers and incorrect ultimate response. Hopefully, there is an option to [process the redirects manually](https://github.com/bigcommerce/stencil-cli/compare/master...MaxGenash:STRF-8705_Bug_local_server_is_broken_after_updating_packages?expand=1#diff-596ce37f7ed6dc6ea50288ac9903be41R81).
4. After we start using the WHATWG URL instead of the deprecated Node's "url" library we got a [typo of using the absent "path" property ](https://github.com/bigcommerce/stencil-cli/compare/master...MaxGenash:STRF-8705_Bug_local_server_is_broken_after_updating_packages?expand=1#diff-4645f2b850a88f2c4bfcf2908ff2fac5L33) in the redirect location header URL.

Covered these bugs with tests.

#### Tickets / Documentation

STRF-8705

### DIFF
--- a/server/plugins/renderer/renderer.module.js
+++ b/server/plugins/renderer/renderer.module.js
@@ -69,14 +69,16 @@ internals.getResponse = async (request) => {
         : new URL(request.app.storeUrl);
 
     const httpOpts = {
-        headers: {
-            ...internals.getHeaders(request, { get_template_file: true, get_data_only: true }),
-            host: staplerUrlObject.host,
-        },
+        headers: internals.buildReqHeaders({
+            request,
+            stencilOptions: { get_template_file: true, get_data_only: true },
+            extraHeaders: { host: staplerUrlObject.host },
+        }),
         // Fetch will break if request body is Stream and server response is redirect,
         //  so we need to read the data first and then send the request
         body: request.payload ? await readFromStream(request.payload) : request.payload,
-        method: 'post',
+        method: request.method,
+        redirect: 'manual',
     };
 
     const url = Object.assign(new URL(request.url.toString()), {
@@ -116,19 +118,14 @@ internals.getResponse = async (request) => {
 
     const response = await fetch(url, httpOpts);
 
+    internals.processResHeaders(response.headers);
+
     if (response.status === 401) {
         return response;
     }
 
     if (response.status === 500) {
         throw new Error('The BigCommerce server responded with a 500 error');
-    }
-
-    if (response.headers.get('set-cookie')) {
-        response.headers.set(
-            'set-cookie',
-            utils.stripDomainFromCookies(response.headers.get('set-cookie')),
-        );
     }
 
     // Response is a redirect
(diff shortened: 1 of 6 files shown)


server/plugins/renderer/renderer.module.js
```diff
@@ -429,7 +419,23 @@ internals.getHeaders = (request, stencilOptions = {}, stencilConfig) => {
...
+ * @param {Headers} headers
+ * @returns {object}
+ */
+internals.processResHeaders = (headers) => {
+    if (headers.get('set-cookie')) {
+        // When there are several values for the same header headers.get() returns
+        //  an array joined with comma which is wrong for cookies, so we use the initial raw Array
+        // https://github.com/node-fetch/node-fetch/issues/251#issuecomment-428143940
+        // eslint-disable-next-line no-param-reassign
+        headers.raw()['set-cookie'] = utils.stripDomainFromCookies(headers.raw()['set-cookie']);
```
jairo-bc, line 437: Maybe it's time to consider another library? :)

jairo-bc (approved):

server/plugins/renderer/renderer.module.js
MaxGenash: Maybe😄 (but in a separate PR)
Which one do you recommend? Axios?

junedkazi (approved):

server/plugins/renderer/renderer.module.js
jairo-bc: yeah, I enjoy working with axios

MaxGenash: @junedkazi if you agree, I'll create a ticket for replacing fetch with Axios since Axios figured out to be more handy

junedkazi: Let's create a ticket and we can switch over to axios.

MaxGenash: @jairo-bc , @junedkazi , ticket: https://jira.bigcommerce.com/browse/STRF-8745

PR #648, opened 2020-09-29
Move tests to GitHub Actions

bookernath:
#### What?

Move tests to GitHub actions, allowing us to eliminate Travis and AppVeyor.

Testing a matrix of Linux/Windows/OSX across Node 10 and 12.

ping @junedkazi

PR #659, opened 2020-10-12
fix: added missed linting step to GitHub actions

MaxGenash:
#### What?

in https://github.com/bigcommerce/stencil-cli/pull/648 we moved from Travis+Appveyour to Github actions, but missed the linting step in the new CI/CD config, so I added it back here.

**Upd**: also added .gitattributes to fix problems with prettier on Windows https://stackoverflow.com/a/55998281/9193658 .

And updated the testing step to check test coverage

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Set a consistent line-ending styles across different OS to avoid problems with linters
+* text eol=lf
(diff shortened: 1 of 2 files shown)


junedkazi (approved):

jairo-bc:
@MaxGenash There is a broken test for windows, should latest commit fix it?

jairo-bc (approved):

junedkazi (approved):

Issue #657, opened 2020-10-09
Images not loaded when using stencil-cli 3.0.1


aarasaki:
### Expected behavior
Image's from loading from webdav folders
![image](https://user-images.githubusercontent.com/40073912/95613732-42bc5b80-0a1a-11eb-8615-3d77bde76d48.png)

### Actual behavior
Images not loading from webdav folders 
<img width="1546" alt="Screen Shot 2020-10-09 at 9 19 57 AM" src="https://user-images.githubusercontent.com/40073912/95613643-1b658e80-0a1a-11eb-8cc1-b4d4c3f494dc.png">

Example of viewing actual image url
![image](https://user-images.githubusercontent.com/40073912/95613556-fffa8380-0a19-11eb-8c42-060fdb63285b.png)

No errors shown in console nor terminal

### Steps to reproduce behavior
npm - v6.14.8
node - v10.17.0
OS - MacOS 10.15.7

MaxGenash:
@aarasaki thanks for the report.
I'll take a look at what may be causing this issue

nguyenchris:
Same issue. SVG file types seem to work fine. Every other image file type seem to produce this issue.

junedkazi:
this should be fixed in the latest cli 3.0.3 version. Please let us know if the issue persists.

PR #661, opened 2020-10-19
fix: (strf-8746) make local server parse binary data responses right

MaxGenash:
#### What?

- After moving from wreck to fetch several PRs ago non-JSON storefront server responses started being treated as text, which resulted in returning binary data (esp. images) in a wrong format (text instead of Buffer). It was fixed here.
- Covered this case with unit tests to avoid such bugs in the future.
- Updated linting npm scripts to ignore unsupported file formats (without it prettier tried to format the content of an image file as a text file).

#### Tickets / Documentation

Ticket: STRF-8746
GitHub Issue: https://github.com/bigcommerce/stencil-cli/issues/657


bookernath (approved):
Thanks for the tests!

Issue #523, opened 2019-10-03
Don't store access credentials with layouts


scottsb:
### Expected behavior
Credentials which should not be tracked in version control should be stored separately from theme configuration which should be.

### Actual behavior
Both credentials (`accessToken`) and configuration (`customLayouts`) are stored in the same file (`.stencil`).

### Steps to reproduce behavior
Use Stencil.

PeteyRev:
@scottsb the `.stencil` file is ignored via `.gitignore`, unsure on what you are referring to?

[https://github.com/bigcommerce/cornerstone](https://github.com/bigcommerce/cornerstone)
```
node_modules/
.stencil
.DS_Store
assets/css-artifacts
assets/js/bundle.js
assets/dist
vendor/bundle
.bundle
*.js.map
*.zip
.idea/
```

MaxGenash:
@scottsb good suggestion, thanks.
I've added a corresponding ticket to our internal board

PR #662, opened 2020-10-29
STRF-8747: split .stencil file into 2 separate config files

MaxGenash:
#### What?

- Moved code dealing with `.stencil` file into a separate class.
- Split `.stencil` file into 2 separate config files: `secrets.stencil.json` (contains "accessToken", "githubToken") and `config.stencil.json` (all other fields), so that themes developers can optionally keep general config data (`config.stencil.json`) in git and add `secrets.stencil.json` to .gitignore. Also implemented migration from an old config file format on config read to avoid breaking changes.
- Increased coverage threshold since we have a higher level at the moment.

#### Tickets / Documentation

Ticket: STRF-8747
Issue: https://github.com/bigcommerce/stencil-cli/issues/523


bookernath:
👏 thanks for doing this.

As some existing users of CLI may be expecting the `.stencil` file to be used, we should probably release this as a new major version, `4.0`.

bookernath:
@MaxGenash can you also add these files to `.gitignore` in Cornerstone?

https://github.com/bigcommerce/cornerstone/blob/master/.gitignore#L2

This will help make sure theme developers do not accidentally commit them to source control.

MaxGenash:
@bookernath I implemented smooth backward compatibility, so at the moment when users reference the `.stencil` file in any CLI command they will receive a message about updating the file, it will be automatically replaced and the users can continue working with CLI, **no breaking changes** yet.
So I believe we can release these changes as v3.1.0 (feature update) and later drop `.stencil` file support in the v4 release

Screenshot 2020-10-30 at 11 17 59

lib/StencilConfigManager.js
```diff
@@ -0,0 +1,134 @@
...
+
+        await this._fs.promises.writeFile(this.configPath, JSON.stringify(generalConfig, null, 2));
+        await this._fs.promises.writeFile(this.secretsPath, JSON.stringify(secretsConfig, null, 2));
+    }
+
+    /**
+     * @private
+     * @param {object} config
+     */
+    _splitStencilConfig(config) {
```
jairo-bc, line 82: maybe its time to use private fields? They are supported from node 12
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

Or do you want to id separately across the project?

lib/stencil-init.js
```diff
@@ -1,75 +1,67 @@
...
-                return this.jsonLint.parse(dotStencilFile, dotStencilFilePath);
-            } catch (err) {
-                this.logger.error(
-                    'Detected a broken .stencil file:\n',
-                    err,
-                    '\nThe file will be rewritten with your answers',
-                );
-            }
+    async readStencilConfig() {
+        let parsedConfig;
```
jairo-bc, line 52: `let parsedConfig = {};`
and then just 
`return parsedConfig`

MaxGenash: there are cases when this._stencilConfigManager.read() returns null (config doesn’t exist) and we should return an empty object in this method instead of null

lib/StencilConfigManager.js
MaxGenash: We still support Node 10 in this project (see https://github.com/bigcommerce/stencil-cli/pull/645#discussion_r492898129) and it doesn’t have support for private fields yet

jairo-bc (approved):

lib/StencilConfigManager.js
jairo-bc: uh, okay, I thought we already dropped the node 10

bookernath:
@MaxGenash great, good thinking and thank you for clarifying!

MaxGenash:
@junedkazi checked, we are not pushing the credentials file when we do a stencil push or when we make a bundle

MaxGenash:
@bookernath @junedkazi before releasing these changes, need to update documentation on developer.bigcommerce.com to replace mentions of .stencil with the new files and update videos on YouTube where .stncil file is edited (e.g. https://youtu.be/qgaDX7bhmd8 )

junedkazi (approved):

Issue #665, opened 2020-11-11
Stencil Start Error 


joynakakihara:
Hi,
I have upgraded from stencil-cli version 1.23.1 to version 3.0.3. Executing stencil start produces the following:
error not ok -- Error: The BigCommerce Store you are pointing to either does not exist or is not available at this time.
Error details:
request to https://spinning.com/stencil-version-check?v=3.0.3 failed, reason: unable to verify the first certificate. 

I have taken the following steps trying to resolve the issue.
1. Uninstall/reinstall stencil-cli
2. Uninstall/reinstall Node.js (currently using Node version 12)
3.Uninstall/reinstall Python 2.7

Still facing the same issue.  Any ideas? 

Thanks,
Joy

MaxGenash:
@joynakakihara  
Seems like there are some problems with SSL certificates.
As a temporary solution, you can use this one https://stackoverflow.com/a/48960988/9193658 (since stencil-cli is used for dev purposes only and we send requests to our servers only - disabling SSL checks here is not dangerous).
Which means:
```
export NODE_TLS_REJECT_UNAUTHORIZED=0 
stencil start   
```
OR
```
NODE_TLS_REJECT_UNAUTHORIZED=0 stencil start
```

I'll investigate what may be causing this issue and create a PR with a fix later.

MaxGenash:
Created a ticket for this issue: STRF-8840

PR #669, opened 2020-12-04
Strf 8745 replace fetch with axios

MaxGenash:
#### What?

- Replaced fetch with axios (STRF-8745).
- Added handling cases when redirect link is already stripped in normalizeRedirectUrl(). Previously I didn’t encounter problems with it, so maybe it's just because axios does it under the hood.
- Added missing rejectUnauthorized parameter to the API requests (strf-8840).

#### Tickets / Documentation

STRF-8840 = https://github.com/bigcommerce/stencil-cli/issues/665
STRF-8745


lib/theme-api-client.js
```diff
@@ -77,38 +65,31 @@ async function deleteThemeById({ themeId, accessToken, apiHost, storeHash }) {
...
-
     try {
-        response = await fetch(reqUrl, reqOpts);
-        payload = await response.json();
+        response = await sendApiRequest({
+            url: `${apiHost}/stores/${storeHash}/v3/themes/jobs/${jobId}`,
+            headers: {
+                'cache-control': 'no-cache',
+                'x-auth-client': 'stencil-cli',
+                'x-auth-token': accessToken,
```
junedkazi, line 76: do we need to add `content-type` header here ? I am wondering if we should remove the headers outside and keep it generic to all calls.

junedkazi (approved):

lib/theme-api-client.js
MaxGenash: @junedkazi 
It's a GET request, so we aren’t sending any data here (no content, so no content-type).

Regarding moving the headers inside `sendApiRequest` - it's a good idea, but there are no headers that we send in all our requests. Though, we can move 
```
{
    'x-auth-client': 'stencil-cli',
    'stencil-cli': PACKAGE_INFO.version,
    'stencil-version': PACKAGE_INFO.config.stencil_version,
}
```
since we are sending them in the majority of the requests and they won’t hinder us if we send them when not required.

junedkazi: ya we should consolidate the generic ones in a single place and avoid duplication

lib/utils/networkUtils.js
```diff
@@ -2,22 +2,36 @@
...
+
+/** Used to send request to our (Bigcommerce) servers only.
+ *  Shouldn't be used to send requests to third party servers because we disable https checks
+ *
+ * @param {object} options
+ * @returns {Promise<object>}
+ */
+async function sendApiRequest(options) {
+    return axios({
+        httpsAgent: new https.Agent({ rejectUnauthorized: false }),
```
jairo-bc, line 17: Do we really need to create a new agent for each request?

MaxGenash: Maybe we can reuse it, kinda:
```
const httpsAgent = new https.Agent({ rejectUnauthorized: false });
axios({ httpsAgent, ...options });
```
But need to check if it doesn’t cache or reuse something leading to unexpected behavior.

MaxGenash: Updated in the new commit

lib/theme-api-client.js
MaxGenash: Updated in the new commit

jairo-bc (approved):

PR #679, opened 2021-01-13
fix:(STRF-8909) theme variation not applied with activate flag

huntario:
#### What?

The --activate flag should allow for the user to specify the theme variation to apply when pushing a theme, e.g. `stencil push --activate Warm` However, the first theme variation is always applied. The variation parameter form the CLI is ignored. 

#### Tickets / Documentation

STRF-8909

BEFORE: The variation specified in the CLI is ignored. The first variation in the list is always applied.

Screen Shot 2021-01-13 at 4 19 54 PM

AFTER: The variation specified in the CLI is applied when using the --activate flag

Screen Shot 2021-01-13 at 4 17 08 PM

### DIFF
--- a/lib/stencil-push.utils.js
+++ b/lib/stencil-push.utils.js
@@ -311,7 +311,7 @@ utils.getVariations = async (options) => {
         return { ...options, variationId: foundVariation.uuid };
     }
     if (activate) {
-        const indexofVariationToActivate = variations.findIndex(variation => variation.name === activate);
+        const indexofVariationToActivate = variations.findIndex((variation) => variation.name === activate);
         if (indexofVariationToActivate === -1) {
             const availableOptionsStr = variations.map((item) => `${item.name}`).join(', ');
             throw new Error(
--- a/lib/stencil-push.utils.js
+++ b/lib/stencil-push.utils.js
@@ -311,7 +311,6 @@ utils.getVariations = async (options) => {
         return { ...options, variationId: foundVariation.uuid };
     }
     if (activate) {
-        // Activate the theme variant with the name matches the CLI argument
         const indexofVariationToActivate = variations.findIndex(variation => variation.name == activate);
         if (indexofVariationToActivate === -1) {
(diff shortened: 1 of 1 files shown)


huntario:
@NikitaPuza @bdav87 @briansilvey @junedkazi Would it be possible to get a review of this when there is some time? Let me know if there is anything that looks worrisome.

lib/stencil-push.utils.js
```diff
@@ -311,7 +311,16 @@ utils.getVariations = async (options) => {
         return { ...options, variationId: foundVariation.uuid };
     }
     if (activate) {
-        return { ...options, variationId: variations[0].uuid };
+        const indexofVariationToActivate = variations.findIndex(
```
jairo-bc, line 314: I think we can simplify to (check the condition above) 
`const foundVariation = variations.find((item) => item.name === activate);`
and 
`return { ...options, variationId: foundVariation.uuid };`

jairo-bc: Also I don't think we need the previous condition 
`if (!activate && activate !== undefined) {`
it will work only if activate is _0_ or _""_ or _false_

jairo-bc (changes requested):

lib/stencil-push.utils.js
MaxGenash: Yeah, I actually suggested these changes in the provided fixed code above)

jairo-bc: @MaxGenash Agree. That's the better way

huntario:
@MaxGenash @jairo-bc I really appreciate y'all taking a look and providing this advice. I've submitted the requested changes.

jairo-bc (approved):

MaxGenash (approved):

MaxGenash:
@huntario I've merged the fix, but the release is blocked by changes in https://github.com/bigcommerce/stencil-cli/pull/662#issuecomment-721703355 (waiting for the updates of the documentation)

PR #680, opened 2021-01-14
fix(stencil-push): (STRF-8913) increase maxBodyLength in NetworkUtils.sendApiRequest

MaxGenash:
#### What?

Noticed that after moving to Axios we got a bug that sometimes `stencil push` command fails with a message "not ok – ThemeUploadError: Request body larger than maxBodyLength limit".

To fix the bug I just increased maxBodyLength and maxContentLength to Infinite values.

#### Tickets / Documentation

STRF-8913

### DIFF
--- a/lib/utils/NetworkUtils.js
+++ b/lib/utils/NetworkUtils.js
@@ -35,6 +35,8 @@ class NetworkUtils {
     async sendApiRequest(options) {
         const { accessToken, ...restOpts } = options;
         const reqConfig = {
+            maxContentLength: Infinity,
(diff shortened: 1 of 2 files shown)


jairo-bc:
Is it 10MB by default?

MaxGenash:
> Is it 10MB by default?

Yes. Actually, axios just passes it to [follow-redirects](https://www.npmjs.com/package/follow-redirects#options) package which has a default value of 10MB.

jairo-bc (approved):

PR #690, opened 2021-03-12
Updated stencil push and pull commands to support multiple channels

MaxGenash:
#### What?

Updated stencil push and pull commands to support multiple channels.

#### Tickets / Documentation

- feat: [STRF-7651](https://jira.bigcommerce.com/browse/STRF-7651) update push command
- feat: [STRF-8282](https://jira.bigcommerce.com/browse/STRF-8282) - update pull command

`stencil push`

- When there is only 1 channel (skips channel selection):
push when only 1 channel

- When we specify channel_id in arguments (skips channel selection):
push with specified channel_id

- When there are several channels available:
stencil push choice
push final

`stencil pull`

- When there is only 1 channel (skips channel selection):
pull when only 1 channel

- When we specify channel_id in arguments (skips channel selection):
pull with specified channel_id

- When there are several channels available:
pull choice
pull final

### DIFF
--- a/bin/stencil-pull.js
+++ b/bin/stencil-pull.js
@@ -30,8 +30,9 @@ const cliOptions = program.opts();
 const options = {
     apiHost: cliOptions.host || API_HOST,
     saveConfigName: cliOptions.filename,
-    channelId: cliOptions.channel_id || 1,
+    channelId: cliOptions.channel_id,
     saved: cliOptions.saved || false,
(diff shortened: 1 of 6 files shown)


junedkazi (approved):

jairo-bc (approved):

MaxGenash:
I updated the PR replacing channels API to sites API because stencil-cli token doesn’t have access to channels API, but has access to sites API

PR #726, opened 2021-06-29
feat(other): MERC-7797 Added feature to render Widgets in the regions of supported page types.

jkanive:
#### What?

Added logic to rendered the widgets in the respective regions of a page. This is done by making GraphQL request for each page request and setting the content via the paper handlebar helper `setContent`.

#### Tickets / Documentation

-   [MERC-7797](https://jiira.bigcommerce.com/browse/MERC-7797)

@junedkazi Would appreciate your input on this PR.

### DIFF
--- a/lib/content-api-client.js
+++ b/lib/content-api-client.js
@@ -1,0 +1,97 @@
+require('colors');
+const NetworkUtils = require('./utils/NetworkUtils');
+const {
+    renderedRegionsByPageTypeQuery,
+    renderedRegionsByPageTypeAndEntityIdQuery,
+} = require('./graphql/query');
+
+const networkUtils = new NetworkUtils();
+
+/**
+ * @param {object} options
+ * @param {string} options.accessToken
+ * @param {string} options.storeUrl
+ * @param {string} pageType
+ * @returns {Promise<renderedRegions: array}>}
+ */
+async function getRenderedRegionsByPageType({ accessToken, storeUrl, pageType }) {
+    try {
+        const query = renderedRegionsByPageTypeQuery(pageType);
+
+        const response = await networkUtils.sendApiRequest({
+            url: `${storeUrl}/graphql`,
+            headers: {
+                'cache-control': 'no-cache',
+                'content-type': 'application/json',
+                Authorization: `Bearer ${accessToken}`,
+            },
+            method: 'POST',
+            data: JSON.stringify({
+                query,
+            }),
+        });
+
+        const {
+            site: {
+                content: {
+                    renderedRegionsByPageType: { regions },
+                },
+            },
+        } = response.data.data;
+
+        return { renderedRegions: regions };
+    } catch (err) {
+        throw new Error(`Could not fetch the rendered regions for this page type: ${err.message}`);
+    }
+}
+
+/**
+ * @param {object} options
+ * @param {string} options.accessToken
+ * @param {string} options.storeUrl
+ * @param {string} pageType
+ * @param {number} entityId
+ * @returns {Promise<renderedRegions: array}>}
+ */
+async function getRenderedRegionsByPageTypeAndEntityId({
+    accessToken,
+    storeUrl,
+    pageType,
+    entityId,
+}) {
+    try {
+        const query = renderedRegionsByPageTypeAndEntityIdQuery(pageType, entityId);
+
+        const response = await networkUtils.sendApiRequest({
+            url: `${storeUrl}/graphql`,
+            headers: {
+                'cache-control': 'no-cache',
+                'content-type': 'application/json',
+                Authorization: `Bearer ${accessToken}`,
+            },
+            method: 'POST',
+            data: JSON.stringify({
+                query,
+            }),
+        });
+
+        const {
+            site: {
+                content: {
+                    renderedRegionsByPageTypeAndEntityId: { regions },
+                },
+            },
+        } = response.data.data;
+
+        return {
+            renderedRegions: regions,
+        };
+    } catch (err) {
+        throw new Error(`Could not fetch the rendered regions for this page type: ${err.message}`);
+    }
+}
+
+module.exports = {
+    getRenderedRegionsByPageType,
+    getRenderedRegionsByPageTypeAndEntityId,
+};
--- a/server/index.js
+++ b/server/index.js
@@ -11,10 +11,10 @@ function buildManifest(srcManifest, options) {
 
     const parsedSecureUrl = new URL(options.dotStencilFile.storeUrl); // The url to a secure page (prompted as login page)
     const parsedNormalUrl = new URL(options.dotStencilFile.normalStoreUrl); // The host url of the homepage;
+    const storeUrl = parsedSecureUrl.protocol + '//' + parsedSecureUrl.host;
 
     resManifest.server.port = options.dotStencilFile.port;
-    pluginsByName['./plugins/router/router.module'].storeUrl =
-        parsedSecureUrl.protocol + '//' + parsedSecureUrl.host;
+    pluginsByName['./plugins/router/router.module'].storeUrl = storeUrl;
     pluginsByName['./plugins/router/router.module'].normalStoreUrl =
         parsedNormalUrl.protocol + '//' + parsedNormalUrl.host;
     pluginsByName['./plugins/router/router.module'].apiKey = options.dotStencilFile.apiKey;
@@ -27,6 +27,7 @@ function buildManifest(srcManifest, options) {
     pluginsByName['./plugins/renderer/renderer.module'].customLayouts =
         options.dotStencilFile.customLayouts;
     pluginsByName['./plugins/renderer/renderer.module'].themePath = options.themePath;
+    pluginsByName['./plugins/renderer/renderer.module'].storeUrl = storeUrl;
     pluginsByName['./plugins/theme-assets/theme-assets.module'].themePath = options.themePath;
 
     resManifest.register.plugins = _.reduce(
--- a/server/lib/page-type-util.js
+++ b/server/lib/page-type-util.js
@@ -1,0 +1,101 @@
+const PageTypes = {
+    PAGE: 'PAGE',
+    PRODUCT: 'PRODUCT',
+    CATEGORY: 'CATEGORY',
+    BRAND: 'BRAND',
+    ACCOUNT_RETURN_SAVED: 'ACCOUNT_RETURN_SAVED',
+    ACCOUNT_ADD_RETURN: 'ACCOUNT_ADD_RETURN',
+    ACCOUNT_RETURNS: 'ACCOUNT_RETURNS',
+    ACCOUNT_ADD_ADDRESS: 'ACCOUNT_ADD_ADDRESS',
+    ACCOUNT_ADD_WISHLIST: 'ACCOUNT_ADD_WISHLIST',
+    ACCOUNT_WISHLISTS: 'ACCOUNT_WISHLISTS',
+    ACCOUNT_WISHLIST_DETAILS: 'ACCOUNT_WISHLIST_DETAILS',
+    ACCOUNT_EDIT: 'ACCOUNT_EDIT',
+    ACCOUNT_ADDRESS: 'ACCOUNT_ADDRESS',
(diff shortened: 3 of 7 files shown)


lib/content-api-client.js
```diff
@@ -0,0 +1,86 @@
...
+                Authorization: `Bearer ${accessToken}`,
+            },
+            method: 'POST',
+            data: JSON.stringify({
+                query,
+            }),
+        });
+
+        return {
+            renderedRegions: response.data.data.site.content.renderedRegionsByPageType.regions,
```
bc-dhuynh, line 35: 😨  can we possibly deconstruct this response?

server/lib/page-type-util.js
```diff
@@ -0,0 +1,182 @@
...
+ * @param {string} templateFile
+ * @returns {string | null}
+ */
+function getPageType(templateFile) {
+    let pageType = null;
+    switch (templateFile) {
+        case 'pages/page':
+            pageType = 'PAGE';
+            break;
+        case 'pages/product':
```
bc-dhuynh, line 13: What do you think about having a mapping object?
```
pageTypes = { 'pages/page': 'PAGE' };

return pageTypes[templateFile];
```

server/lib/page-type-util.js
```diff
@@ -0,0 +1,182 @@
...
+    let entityRequired = false;
+
+    if (typeof entityId !== 'number') {
+        return false;
+    }
+
+    switch (pageType) {
+        case 'BLOG_POST':
+            entityRequired = true;
+            break;
```
bc-dhuynh, line 155: same as my question above

server/plugins/renderer/renderer.module.js
```diff
@@ -198,7 +201,53 @@ internals.parseResponse = async (bcAppData, request, response, responseArgs) =>
...
     }
 
-    return internals.getPencilResponse(response2.data, request, response2, configuration);
+    const templateFile = response2.data.template_file;
+    const entityId = response2.data.entity_id;
+
+    const pageType = getPageType(templateFile);
+    let regionResponse = [];
+
+    if (pageType !== null) {
```
bc-dhuynh, line 210: `if (pageType) { ... }`?

What are the possibilities here?

server/plugins/renderer/renderer.module.js
```diff
@@ -198,7 +201,53 @@ internals.parseResponse = async (bcAppData, request, response, responseArgs) =>
...
+        // create request signature and use cache, if available
+        const graphQLUrlSignature = internals.sha1sum(internals.options.storeUrl + '/graphql');
+        const graphQLQuerySignature = internals.sha1sum(pageType + entityId);
+        const graphQLDataReqSignature = `graphql:${graphQLUrlSignature}${graphQLQuerySignature}`;
```
bc-dhuynh, line 214: `graphql:${graphQLUrlSignature + graphQLQuerySignature}`?

server/plugins/renderer/renderer.module.js
```diff
@@ -198,7 +201,53 @@ internals.parseResponse = async (bcAppData, request, response, responseArgs) =>
...
+        const cachedGraphQLResponse = cache.get(graphQLDataReqSignature);
+
+        if (internals.options.useCache && cachedGraphQLResponse) {
+            ({ regionResponse } = cachedGraphQLResponse);
```
bc-dhuynh, line 218: What does this mean?

server/plugins/renderer/renderer.module.js
jkanive: I am setting the value to `null` specifically in `getPageType`. But I can see using just `if (pageType) { ... }`.

server/plugins/renderer/renderer.module.js
jkanive: It's extracting cached value of `regionResponse` instead of making a graphQL call.

lib/content-api-client.js
jkanive: I am not sure if deconstruction would help as there is nothing else we need to extract except for regions 🤷‍♂️

jkanive:
♻️ Addressed comments.

lib/content-api-client.js
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * @param {object} options
+ * @param {string} options.accessToken
+ * @param {string} options.storeUrl
+ * @param {string} pageType
+ * @param {number} entityId
+ * @returns {Promise<renderedRegions: array}>}
+ */
+async function getRenderedRegionsByPageTypeAndEntityId({
```
christensenep, line 58: Kinda surprised these need to be two separate functions, isn't really just affecting the query slightly when you don't supply an entity id?

christensenep: Oh wait right as I clicked the button I realized that parsing the response is slightly different too, yeah fair enough.

server/lib/page-type-util.js
```diff
@@ -0,0 +1,126 @@
...
+    GIFT_CERT_REDEEM: 'GIFT_CERT_REDEEM',
+    GIFT_CERT_BALANCE: 'GIFT_CERT_BALANCE',
+    ORDER_INFO: 'ORDER_INFO',
+    SEARCH: 'SEARCH',
+    SITEMAP: 'SITEMAP',
+    SUBSCRIBED: 'SUBSCRIBED',
+    UNSUBSCRIBE: 'UNSUBSCRIBE',
+};
+
+const templateFileToPageTypeMap = {
```
christensenep, line 44: Really wish we could fetch this from somewhere or include it as a library or something.

jkanive: This mapping is in Storefront also now, so we technically are doing this in 3 places. BCApp, Storefront and now in Stencil CLI 🤷‍♂️

server/lib/page-type-util.js
```diff
@@ -0,0 +1,126 @@
...
+ * @param {string} pageType
+ * @param {number} entityId
+ * @returns {boolean}
+ */
+function isValidPageTypeEntityIdReq(pageType, entityId) {
+    if (typeof entityId !== 'number') {
+        return false;
+    }
+
+    return pageTypeWithEntityId.indexOf(pageType) >= 0;
```
christensenep, line 120: Is this actually necessary? Can we not just check if the entityId is present and just trust that it's correct rather than checking another list we have to maintain here?

server/plugins/renderer/responses/pencil-response.js
```diff
@@ -137,6 +138,7 @@ class PencilResponse {
         this.data.context.in_production = false;
 
         paper.addDecorator(makeDecorator(request, this.data.context));
+        paper.setContent(this.data.renderedRegions);
```
christensenep, line 141: What?!? My mind is blown.

lib/content-api-client.js
jkanive: When using the GraphQL Playground to build the query it pretty handy because it does the validation of whether a pageType needs entityId or not. I am guessing that was the reasoning for having separate function as its difficult to make a argument conditionally required  on the same function.

server/plugins/renderer/renderer.module.js
```diff
@@ -12,12 +12,15 @@ const templateAssembler = require('../../../lib/template-assembler');
...
 const NetworkUtils = require('../../../lib/utils/NetworkUtils');
+const contentApiClient = require('../../../lib/content-api-client');
+const { getPageType, isValidPageTypeEntityIdReq } = require('../../lib/page-type-util');
 
 const networkUtils = new NetworkUtils();
 
 const internals = {
     options: {},
     cacheTTL: 1000 * 15, // 15 seconds
+    graphQLCacheTTL: 1000 * 300, // 5 minutes
```
christensenep, line 23: Just a thought for the future, it might be worth considering literally never refreshing the cache without user intervention, or making it a cli setting or something. Just imagining in the vast majority of use cases the user wouldn't be expecting or needing the widgets to update. (I don't really have a sense for how many users are using stencil cli at any given time, admittedly it's probably only a handful so not a huge deal at the moment)

server/plugins/renderer/responses/pencil-response.js
jkanive: We have paper helper which we use in Storefront renderer as well which takes input of an object with keys as the regionName and value as the html. So I am formatting it and just feeding it here. I was doing literal search a replace before I thought of figuring out how storefront renderer does it.

lib/content-api-client.js
christensenep: Oh sorry I was speaking of *these* functions, not the graphql query itself. But anyway yeah given that it's a different graphql field think this makes sense.

server/plugins/renderer/responses/pencil-response.js
christensenep: Oh yeah sorry that was not actually a question, I was just amazed how simple it was. I know you told me before about this, just seeing it is kind of amazing.

server/plugins/renderer/renderer.module.js
jkanive: There is an option to start the stencil-cli without using cache `stencil start --no-cache`. So it might be work changing the TTL to say 60mins or more. I think the TTL is not optional and would throw an error if not provided.

jkanive:
♻️ Addressed comments.

christensenep (approved):

lib/content-api-client.js
```diff
@@ -0,0 +1,99 @@
...
+                Authorization: `Bearer ${accessToken}`,
+            },
+            method: 'POST',
+            data: JSON.stringify({
+                query,
+            }),
+        });
+
+        const {
+            site: { content },
```
jairo-bc, line 35: is `content` variable for readability? 
It's possible to do like this: 
```suggestion
            const { site: { regions: { renderedRegionsByPageType } } } = response.data.data
```

lib/content-api-client.js
```diff
@@ -0,0 +1,99 @@
...
+        } = response.data.data;
+        const {
+            renderedRegionsByPageType: { regions },
+        } = content;
+
+        return {
+            renderedRegions: regions,
+        };
+    } catch (err) {
+        console.log(err);
```
jairo-bc, line 45: `console.error`

lib/content-api-client.js
```diff
@@ -0,0 +1,99 @@
...
+        } = response.data.data;
+        const {
+            renderedRegionsByPageTypeAndEntityId: { regions },
+        } = content;
+
+        return {
+            renderedRegions: regions,
+        };
+    } catch (err) {
+        console.log(err);
```
jairo-bc, line 91: `console.error`
It's also possible to highlight error with red color like here https://github.com/bigcommerce/stencil-cli/blob/master/lib/release/release.js#L145

server/index.js
```diff
@@ -27,6 +27,8 @@ function buildManifest(srcManifest, options) {
     pluginsByName['./plugins/renderer/renderer.module'].customLayouts =
         options.dotStencilFile.customLayouts;
     pluginsByName['./plugins/renderer/renderer.module'].themePath = options.themePath;
+    pluginsByName['./plugins/renderer/renderer.module'].storeUrl =
+        parsedSecureUrl.protocol + '//' + parsedSecureUrl.host;
```
jairo-bc, line 31: Please create a variable `storeUrl` and reuse it on line 16 as well

server/lib/page-type-util.js
jairo-bc: Are we in hurry with this ticket, so we are introducing here tech debt? If so, lets create a ticket to address it by adding an endpoint to Storefront (or other more relevant service)

jairo-bc (changes requested):

jkanive:
> Are we in hurry with this ticket, so we are introducing here tech debt? If so, lets create a ticket to address it by adding an endpoint to Storefront (or other more relevant service)

@jairo-bc I have created a ticket [MERC-7975](https://jira.bigcommerce.com/browse/MERC-7975) to address the mapping tech debt and will look into it in a follow up PR.

jkanive:
♻️ Addressed all the comments.
cc: @jairo-bc @christensenep

jairo-bc (approved):
LGTM

PR #739, opened 2021-07-26
feat: strf-9258 Stencil Pull: Provide an activate option

jairo-bc:
#### What?

Refactored Stencil Pull command to have ability write unit tests.
Added activate option for Stencil Pull

#### Tickets / Documentation

-   [STRF-9258](https://jira.bigcommerce.com/browse/STRF-9258)


zvuki (approved):

lib/stencil-pull.js
```diff
@@ -1,21 +1,208 @@
...
+        console.log('ok'.green + ` -- Fetched theme details for channel ${channelId}`);
+
+        const variations = await this._themeApiClient.getVariationsByThemeId({
+            accessToken: stencilConfig.accessToken,
+            apiHost: cliOptions.apiHost,
+            themeId: activeTheme.active_theme_uuid,
+            storeHash,
+        });
+
+        const variationId = this._stencilPushUtils.getActivatedVariation(
```
mattolson, line 61: What happens if this returns null?

mattolson (commented):
Code looks good, but not sure about the naming of the option. If I understand correctly, the new option overrides the variation to use when fetching the configuration to merge. So it doesn't really "activate" a variation, it merges the given variation's configuration (as opposed to the default variation).

Thinking about this functionality more, I'm not sure it's working in the best possible way. When we merge these configurations, we put it into the top-level "settings" object. However, the variations still exist in the config file, and when this is re-uploaded, those variations will be layered on top of the base settings object. So if you think about the full experience of `stencil pull` + edits + `stencil push` against the Cornerstone Bold variations for example, we would pull the existing Bold configuration, merge it into the default settings object, but because the Bold variation still exists in config.json, when re-uploaded, we wouldn't see those settings, we would see whatever is still leftover in the Bold variation (a separate object in config.json).

jairo-bc:
@mattolson Maybe the name is not the best. I'm using the same as we currently have for `stencil push` (https://github.com/bigcommerce/stencil-cli/blob/master/bin/stencil-push.js)

Regarding functional approach.
Currently if you run `stencil push`, it will update only global settings. If you run `stencil push --activate Bold`, it will update only Bold variation.
If you run `stencil pull`, it will update only global settings. 

And I'm adding **activate** option for `stencil pull` to have consistency with `stencil push`

PR #744, opened 2021-08-04
fix: (MERC-8038) Added check to ensure we are not looping over undefined object

jkanive:
#### What?

#### Tickets / Documentation
[MERC-8038](https://jira.bigcommerce.com/browse/MERC-8038)

When we are looking at pages that are not supported by the GraphQL renderedRegions field we were getting back an `undefined` value. This was not accounted for when we were looping over the result. This PR fixes that issue.

### DIFF
--- a/server/plugins/renderer/renderer.module.js
+++ b/server/plugins/renderer/renderer.module.js
@@ -237,9 +237,12 @@ internals.parseResponse = async (bcAppData, request, response, responseArgs) => 
     }
 
     const formattedRegions = {};
-    regionResponse.renderedRegions.forEach((region) => {
-        formattedRegions[region.name] = region.html;
-    });
+
+    if (typeof regionResponse.renderedRegions !== 'undefined') {
+        regionResponse.renderedRegions.forEach((region) => {
+            formattedRegions[region.name] = region.html;
+        });
+    }
 
     return internals.getPencilResponse(
         response2.data,


server/plugins/renderer/renderer.module.js
```diff
@@ -237,9 +237,12 @@ internals.parseResponse = async (bcAppData, request, response, responseArgs) =>
     }
 
     const formattedRegions = {};
-    regionResponse.renderedRegions.forEach((region) => {
-        formattedRegions[region.name] = region.html;
-    });
+
+    if (typeof regionResponse.renderedRegions !== 'undefined') {
```
jairo-bc, line 241: I would recommend to use 
```suggestion
    if (!regionResponse.renderedRegions) {
```

in case if _**regionResponse.renderedRegions**_ comes as **null** from backend

jairo-bc (approved):

server/plugins/renderer/renderer.module.js
jkanive: I generally prefer doing explicit checks as its safer and is not prone to errors like something being a boolean. Let me know if you feel strongly about this change request.

jairo-bc: I'm good with the changes. I approved your PR :) 

Just curious what kind of errors are you referring to?

jkanive: In this particular case the `renderedRegions` will never be boolean, but if it were then `if (regionResponse.renderedRegions)` would fail to do the right thing because `false` is a valid value.

PR #754, opened 2021-09-07
feat: strf-9345 Infer apiHost from storeUrl

jmwiese:
[STRF-9345](https://jira.bigcommerce.com/browse/STRF-9345)

#### What?

Attempt to infer the apiHost from the storeUrl for integration/staging stores, falling back to API_HOST constant otherwise.

apiHost is now saved in the stencil config and loaded from there instead of just the API_HOST

@jairo-bc

### DIFF
--- a/bin/stencil-download.js
+++ b/bin/stencil-download.js
@@ -4,14 +4,14 @@ require('colors');
 const inquirer = require('inquirer');
 const program = require('../lib/commander');
 
-const { API_HOST, PACKAGE_INFO } = require('../constants');
+const { PACKAGE_INFO } = require('../constants');
 const stencilDownload = require('../lib/stencil-download');
 const { checkNodeVersion } = require('../lib/cliCommon');
 const { printCliResultErrorAndExit } = require('../lib/cliCommon');
 
 program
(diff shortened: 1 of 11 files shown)


bookernath:
Instead of doing this implicitly, why not just accept an argument on `stencil init` to indicate the environment? Default would be prod, naturally, if not specified.

jmwiese:
@bookernath yeah I can add that, or `--host` can be provided during `stencil start` which is how it works currently. However if someone is trying to test against an integration or staging store (the majority of which will just have a my-integration/my-staging.zone url), why not also try to infer? Because otherwise it will never work.

jairo-bc:
I would suggest to have both `--host` option and inferring, so it will be backward compatible with what we have now. But in this case we need to decide what priority should be.

jmwiese:
@jairo-bc yeah this PR should allow both. It will save the inferred host to the config file. Then during operations like `stencil start`, etc it should prioritize the `--host` option before falling back to the apiHost saved in the config file.

jairo-bc (approved):

jairo-bc:
LGTM

PR #785, opened 2021-10-02
feat: Add overwrite option to download command

jonassvalin:
#### What?
I want the ability to run `stencil download` without being prompted about whether to overwrite local files. The goal is to easier be able to run download in my CI pipeline.

This is a conceptual pull requests thus far, I have not verified whether the functionality works exactly as intended. This particular command seemed to be untested, so I did not venture to add any tests.

If you have preferences for how this functionality could work different, please let me know or feel free to make changes.

### DIFF
--- a/bin/stencil-download.js
+++ b/bin/stencil-download.js
@@ -34,16 +34,16 @@ const options = {
 async function run(opts) {
     const overwriteType = opts.file ? opts.file : 'files';
 
-    const answers =
-        opts.overwrite ||
-        (await inquirer.prompt([
-            {
-                message: `${'Warning'.yellow} -- overwrite local with remote ${overwriteType}?`,
-                name: 'overwrite',
-                type: 'checkbox',
-                choices: ['Yes', 'No'],
-            },
-        ]));
+    const answers = opts.overwrite
+        ? [opts.overwrite]
+        : await inquirer.prompt([
+              {
+                  message: `${'Warning'.yellow} -- overwrite local with remote ${overwriteType}?`,
+                  name: 'overwrite',
+                  type: 'checkbox',
+                  choices: ['Yes', 'No'],
+              },
+          ]);
 
     if (!answers.overwrite.includes('Yes')) {
         console.log(`Request cancelled by user ${'No'.red}`);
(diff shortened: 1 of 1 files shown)


bin/stencil-download.js
```diff
@@ -25,21 +26,24 @@ const options = {
     exclude: ['parsed', 'manifest.json', ...extraExclude],
     apiHost: cliOptions.host,
     channelId: cliOptions.channel_id,
+    overwrite: cliOptions.overwrite,
```
jairo-bc, line 29: If this is just a boolean, I would prefer more specific name, like `skipOverwriteQuestion` or `omitOverwriteAsk`.

bin/stencil-download.js
```diff
@@ -25,21 +26,24 @@ const options = {
...
-            name: 'overwrite',
-            type: 'checkbox',
-            choices: ['Yes', 'No'],
-        },
-    ]);
+    const answers = opts.overwrite
+        ? [opts.overwrite]
+        : await inquirer.prompt([
+              {
+                  message: `${'Warning'.yellow} -- overwrite local with remote ${overwriteType}?`,
```
jairo-bc, line 41: You can use `when` option to disable/enable question: https://github.com/SBoudrias/Inquirer.js#question

jairo-bc (changes requested):
Would be nice to have those comments addressed. Thank you!

bin/stencil-download.js
jonassvalin: So this one isn't actually saying whether to ask the question, but is the actual answer to the question: Should we overwrite? Hence why it seems correct for it to just be called overwrite

bin/stencil-download.js
jonassvalin: Thanks for bringing this to my attention! I have refactored it accordingly.

jonassvalin:
@jairo-bc Thanks for your comments! I have tried to address them as best as I can. 

I made an additional change, which is that I refactored the "overwrite question" to be a confirm, rather than a checkbox. It seemed strange that it was a checkbox since, if my understanding is correct, it's actually just a boolean question, "Should we overwrite the local files?". 

The only downside I see with that change is that it will be a breaking change of how that prompt behaves.

bin/stencil-download.js
jonassvalin: I have refactored this to simply be a boolean, rather than a "Yes" or "No" string value

PR #802, opened 2021-12-02
feat: strf-9245 Warn user if shopper default language file is missing or has missing keys

jairo-bc:
#### What?
As a Stencil CLI user,
When I run `stencil start` for a particular store,
I will receive a warning if the store's default shopper language does not have a language file within the theme.

#### Tickets / Documentation

-   [STRF-9245](https://jira.bigcommerce.com/browse/STRF-9245)

Screenshot 2021-12-02 at 17 24 40

### DIFF
--- a/lib/lang-helper.js
+++ b/lib/lang-helper.js
@@ -1,0 +1,59 @@
+require('colors');
+const fsUtilsModule = require('./utils/fsUtils');
+
+function flattenObject(object, result = {}, parentKey = '') {
+    return Object.entries(object).reduce((currentLayer, [key, innerValue]) => {
+        const resultKey = parentKey !== '' ? `${parentKey}.${key}` : key;
+        if (typeof innerValue === 'object') {
+            return flattenObject(innerValue, currentLayer, resultKey);
+        }
+        // eslint-disable-next-line no-param-reassign
+        currentLayer[resultKey] = innerValue;
+        return currentLayer;
+    }, result);
+}
+
+class LangHelper {
+    constructor({ fsUtils = fsUtilsModule, logger = console } = {}) {
+        this._fsUtils = fsUtils;
+        this._logger = logger;
+    }
+
+    async checkLangKeysPresence(filesPaths, themeLang) {
+        const themeLangFilename = `${themeLang}.json`;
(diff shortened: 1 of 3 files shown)


jmwiese:
I don't quite understand the error messages here. Is the goal to inspect every key for presence? I thought we just wanted to check the ShopperDefaultLanguage, such as `es`, and determine if they have an `es.json` file at all? Because we don't' have a requirement that every translation key is present in any given language file, as long as they are fine with anything that is missing falling back to english.

lib/stencil-start.js
```diff
@@ -236,6 +269,23 @@ class StencilStart {
...
+        const filesPaths = await this._fsUtils.recursiveReadDir(langsPath);
+        const isDefaultLanguagePresent = filesPaths.some((file) =>
+            file.includes(defaultShopperLanguage),
+        );
+
+        if (!isDefaultLanguagePresent) {
+            this._logger.log(
+                `${
+                    'Warning'.yellow
+                }: lang folder is missing default shopper language: ${defaultShopperLanguage}`,
```
jmwiese, line 282: 🍹 I would maybe reword this to be something like "missing language file for default shopper language: ${defaultShopperLanguage}". I think that reads more naturally to me.

lib/lang-helper.js
```diff
@@ -0,0 +1,59 @@
...
+                const langFile = await this._fsUtils.parseJsonFile(filePath);
+                const flattenedLang = flattenObject(langFile);
+                const langKeys = Object.keys(flattenedLang);
+                for (const langKey of langKeys) {
+                    if (!alreadyWarnedKey.includes(langKey)) {
+                        if (!flattenedDefaultLang[langKey]) {
+                            this._logger.log(
+                                `${
+                                    'Warning'.yellow
+                                }: file: ${defaultLangFilePath} doesn't have ${langKey}, which is presented in ${filePath}`,
```
jmwiese, line 42: 🍹 which is presented in -> which is present in

jmwiese (approved):

bookernath:
Nice work!

bookernath (approved):

PR #808, opened 2021-12-10
fix: strf-9553 bundled lang.json has lowercase lang keys

jairo-bc:
#### What?

During bundle process parsed/lang.json now will have lang keys in lowercase.

#### Tickets / Documentation

-   [STRF-9553](https://jira.bigcommerce.com/browse/STRF-9553)

https://user-images.githubusercontent.com/68893868/145610831-ae9d198c-b640-4edf-900e-5c7e7864a205.mov

### DIFF
--- a/lib/lang-assembler.js
+++ b/lib/lang-assembler.js
@@ -26,7 +26,7 @@ function assemble(callback) {
         for (const localeFile of filteredLocaleFiles) {
             const localeName = path.basename(localeFile, '.json');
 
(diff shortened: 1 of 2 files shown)


jmwiese (approved):

jmwiese:
I think this is a good change but we will probably still need another step within renderer to lower case these. Otherwise in order to get the right language behavior every merchant would need to rebundle and reupload their theme which isnt likely to happen.

PR #825, opened 2022-01-05
feat(storefront): strf-9582 stencil push: apply theme to multiple storefronts

bc-max:
#### What?

This feature allows users to apply one theme to multiple storefronts/channels. There are several options available for users:

- `stencil push` command will check whether multiple storefronts/channels available. If yes, user will be prompted to chose one or several/all ids (list of available channels will be shown in radio button style);
- `stencil push -a -c 123 456` allows to apply a theme to selected channels and further system questions will be skipped;
- `stencil push -a -allc` allows to apply a theme to all available channels. 

#### Tickets / Documentation

[STRF-9582](https://jira.bigcommerce.com/browse/STRF-9582)

image

image

image

### DIFF
--- a/lib/theme-api-client.js
+++ b/lib/theme-api-client.js
@@ -69,7 +69,7 @@ async function checkCliVersion({ storeUrl, currentCliVersion = PACKAGE_INFO.vers
  * @param {string} options.apiHost
  * @param {string} options.storeHash
  * @param {string} options.accessToken
- * @returns {Promise<any>}
+ * @returns {Promise<array<any>>}
  */
 async function activateThemeByVariationId({
     variationId,
--- a/lib/theme-api-client.js
+++ b/lib/theme-api-client.js
@@ -65,7 +65,7 @@ async function checkCliVersion({ storeUrl, currentCliVersion = PACKAGE_INFO.vers
 /**
  * @param {object} options
  * @param {string} options.variationId
- * @param {string} options.channelId
+ * @param {array} options.channelIds
  * @param {string} options.apiHost
  * @param {string} options.storeHash
  * @param {string} options.accessToken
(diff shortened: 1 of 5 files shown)


zvuki:
@bc-max It would make sense to have a separate flag, like `stencil push -a -allc` for applying to all channels (so `-c` always expects comma-separated ids as an input).

lib/theme-api-client.js
```diff
@@ -65,33 +65,35 @@ async function checkCliVersion({ storeUrl, currentCliVersion = PACKAGE_INFO.vers
 /**
  * @param {object} options
  * @param {string} options.variationId
- * @param {string} options.channelId
+ * @param {array} options.channelIds
  * @param {string} options.apiHost
  * @param {string} options.storeHash
  * @param {string} options.accessToken
  * @returns {Promise<any>}
```
jairo-bc, line 72: You are now returning array of promises. 
Rather use this function with Promise.all() (which I would recommend) or [for await .. of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of)

jairo-bc (changes requested):

lib/theme-api-client.js
bc-max: I refactored that part. Please take a look.

jairo-bc (approved):

dmaurya-gea:
Might be a dumb question, but does this just assume which variation is being applied? I don't see a variation being passed in the command

jairo-bc:
@dmaurya-gea There's `-a` option for variation name https://github.com/bigcommerce/stencil-cli/blob/master/bin/stencil-push.js#L14

dmaurya-gea:
> @dmaurya-gea There's `-a` option for variation name https://github.com/bigcommerce/stencil-cli/blob/master/bin/stencil-push.js#L14

Yes, I'm aware of that flag and we use it. I saw the same being used here for a different purpose so wanted to check. I wish document had an example with the channels flag

Anyway, I added the `-c` option without the `-a` option while retaining it for the variation name and it worked

`stencil push -a <Variation> -d -c 1` for applying theme to the default channel. For some reason the only storefront was being listed as 2 channels and creating a prompt which broke the build since channel option wasn't passed initially

PR #830, opened 2022-01-11
fix(storefront): strf 9587 api host command is not recognized

bc-max:
#### What?

Terminal doesn't recognise `--apiHost` option when user tries to include it during `stencil init` command

#### Tickets / Documentation

-   [STRF-9587](https://jira.bigcommerce.com/browse/STRF-9587)

image


jmwiese (approved):

lib/stencil-init.spec.js
```diff
@@ -69,12 +77,11 @@ afterEach(() => jest.restoreAllMocks());
 
 describe('StencilInit integration tests', () => {
     describe('run', () => {
-        it('using cli prompts, should perform all the actions, save the result and inform the user about the successful finish', async () => {
+        it('using answers, should perform all the actions, save the result and inform the user about the successful finish', async () => {
```
jairo-bc, line 80: I would rather leave `cli prompts`. That's basically, what the test does.

PR #845, opened 2022-01-20
fix(storefront): BCTHEME-1000 handle correctly css extension in stencil

bc-alexsaiannyi:
#### What?

While working on Prelude POC we would like to change new Stencil Theme structure:
- keep SCSS directory under assets but generate css files under `dist/css` dir,
- use regular `css` extension for it.
- use `css_compiler` flag

If we use mentioned structure, `stencil-cli` gives us an error in local development. For this reason we would like to add possibility for `stencil-cli` to work  with  regular css as well as `aseets/scss` directory. On theme side, we will use `css_compiler` flag  from `config.json` that can be switched between `scss` and `css`. On Stencil side we will provide this extension to `CSS Compiler Handler` which then will assemble CSS accordingly.
 

#### Tickets / Documentation
-   [BCTHEME-1000](https://jira.bigcommerce.com/browse/BCTHEME-1000)

new_assets_dir

https://user-images.githubusercontent.com/67792608/152140829-20e241d6-da21-42e9-ba8a-d7070657c5a2.mov

https://user-images.githubusercontent.com/67792608/152141281-d87c6429-4421-4f10-816f-969c40097ed6.mov

Here our Theme structure at the moment

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,4 @@
 # Set a consistent line-ending styles across different OS to avoid problems with linters
 * text eol=lf
+*.png binary
+*.jpg binary
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### Draft
+
+-   fix(storefront): bctheme-1000 handle regular css in stencil ([845](https://github.com/bigcommerce/stencil-cli/pull/845))
+
 ### 3.9.2 (2022-01-31)
 
 -   chore: bump paper to latest ([863](https://github.com/bigcommerce/stencil-cli/pull/863))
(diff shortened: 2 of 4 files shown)


jairo-bc:
@bc-alexsaiannyi Could you please fix conflict and provide some screenshots as a proof, that is working on cornerstone as well

jairo-bc (commented):
Overall looks good. Take a look at the previous comment

jairo-bc:
I would also recommend to check other commands if they are affected it particular: stencil download and stencil push
@bc-alexsaiannyi

jairo-bc (approved):

Issue #848, opened 2022-01-21
not ok -- VariationActivationError: channelIds.map is not a function


joelzerner:
### Expected behavior

The theme will be pushed up and applied to my Big Commerce instance

### Actual behavior

I see this error:
not ok -- VariationActivationError: channelIds.map is not a function

### Steps to reproduce behavior

stencil push
? Would you like to apply your theme to your store? Yes
? Which variation would you like to apply? Main
not ok -- VariationActivationError: channelIds.map is not a function

### Environment

Stencil-cli version `stencil --version`:
3.8.5

Node version `node -v`:
v12.22.9

NPM version `npm -v`:
6.14.15

OS:
MacOS 11.6

AndrewBarber:
Hey @joelzerner,

It looks like stencil-push was recently changed to support multiple channel push, however, there seems to be a bug when there is only one active 'site'.

One way to get around it at the moment is to specify the channel, the default is always `1` 

So your command would be `stencil push ... -c 1`

matthewvolk:
Hi @joelzerner + @AndrewBarber! 

We just released `@bigcommerce/stencil-cli@3.9.0` which should resolve this issue. When you have a chance, could you update your local Stencil CLI version and see if you can push without explicitly specifying the Channel ID (`-c 1`)?

Thank you @AndrewBarber for the workaround! 

cc: @bc-hnbarr

joelzerner:
Thanks!

I've updated to v3.9.0 and that has resolved the issue.

PR #849, opened 2022-01-21
fix: return channel id as an iterable when a store only has a single storefront

matthewvolk:
#### What?

After adding the ability to use `stencil-push` to push themes to multiple storefronts, users began reporting an error `not ok -- VariationActivationError: channelIds.map is not a function`

This PR includes a fix to return an iterable from the `promptUserToSelectChannel` method even when a store only has a single storefront. 

#### Tickets / Documentation

- #848 

Screen Shot 2022-01-21 at 12 03 57 PM
Screen Shot 2022-01-21 at 12 05 20 PM

### DIFF
--- a/lib/stencil-push.utils.spec.js
+++ b/lib/stencil-push.utils.spec.js
@@ -1,7 +1,12 @@
 const axios = require('axios');
 const MockAdapter = require('axios-mock-adapter');
 
-const { getStoreHash, promptUserForChannel, getChannels } = require('./stencil-push.utils');
+const {
+    getStoreHash,
+    promptUserForChannel,
+    promptUserToSelectChannel,
+    getChannels,
+} = require('./stencil-push.utils');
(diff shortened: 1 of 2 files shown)


lib/stencil-push.utils.spec.js
```diff
@@ -158,4 +163,24 @@ describe('stencil push utils', () => {
             expect(utilsPromptUserForChannelStub).toHaveBeenCalled();
         });
     });
+
+    describe('promptUserToSelectChannel', () => {
+        it('should return an array with a single channel ID if a store only has a single channel', async () => {
```
jairo-bc, line 168: Do you mind moving this test case to the previous suite and rename it to `promptUserToSelectChannel` (so it will follow exactly the function name) ?

jairo-bc (approved):
LGTM

PR #865, opened 2022-02-01
build(release): v3.9.3 dump stencil styles

jairo-bc:
#### What?

Bump stencil styles to support node 14 

### DIFF
--- a/.github/workflows/pull_request_review.yml
+++ b/.github/workflows/pull_request_review.yml
@@ -13,7 +13,7 @@ jobs:
     build:
         strategy:
             matrix:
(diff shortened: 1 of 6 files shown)


.github/workflows/pull_request_review.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     build:
         strategy:
             matrix:
-                node: [10.x, 12.x]
```
jairo-bc, line 16: Are you okay to drop Node 10? @bookernath

bc-max (approved):

PR #874, opened 2022-02-16
add note about npm version to readme

rowleyaj:
#### What?

there are a number of issues that report being unable to install with newer npm versions. the config in package.json only produces a warning when running with a newer npm version, which is drowned out by deprecation notice for other packages. adding it to the readme at least puts a reference where people would typically start.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ _Note: Stencil requires the Node.js runtime environment,
 version 10.x and 12.x (Recommended) are supported.
 We do not yet have support for versions greater than Node 12._
 
-_Review package.json for the required npm version, <7.0.0 at the time of writing_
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,6 +11,8 @@ _Note: Stencil requires the Node.js runtime environment,
 version 10.x and 12.x (Recommended) are supported.
 We do not yet have support for versions greater than Node 12._
 
+_Review package.json for the required npm version, <7.0.0 at the time of writing_
```
jairo-bc, line 14: Do you mind update your PR also with the lowest npm boundary? 
Currently we support node 10.17 and it comes with `npm 6.11.3`

rowleyaj: @jairo-bc thats been done now

rowleyaj:
@jairo-bc I know it's been a while so closing it makes sense. How can we improve the contributor experience though so changes actually get merged? It's a BC project so I appreciate there are internal things happening that customers and developers don't know about however it would be great to get some direction on this?

jairo-bc:
@rowleyaj Be free to ping directly maintainers of the repo (not sure if the team is visible outside of organisation)

PR #878, opened 2022-02-23
fix: when there's no regions on the page

jairo-bc:
#### What?

Sometimes, when new theme is applied to the store, I don't receive `regions` object from the API.

Screenshot 2022-02-18 at 17 57 12

cc @bigcommerce/storefront-team, @jkanive

### DIFF
--- a/lib/content-api-client.js
+++ b/lib/content-api-client.js
@@ -31,6 +31,9 @@ async function getRenderedRegionsByPageType({ accessToken, storeUrl, pageType })
             }),
         });
 
+        if (!response.data.data) {
+            return  { renderedRegions: []};
+        }
         const {
             site: {
                 content: {
@@ -75,6 +78,9 @@ async function getRenderedRegionsByPageTypeAndEntityId({
             }),
         });
 
+        if (!response.data.data) {
+            return  { renderedRegions: []};
+        }
         const {
             site: {
                 content: {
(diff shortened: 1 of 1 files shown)


lib/content-api-client.js
```diff
@@ -31,6 +31,9 @@ async function getRenderedRegionsByPageType({ accessToken, storeUrl, pageType })
             }),
         });
 
+        if (!response.data.data) {
+            return  { renderedRegions: []};
```
jmwiese, line 35: This would probably fix the problem from a stencil-cli perspective, but do we know if there a more root cause as to why this `regions` object isnt populating? Is it just not being passed in properly?

jkanive: I had the same exact comment in the slack. I feel the issue is present somewhere else. `getRenderedRegionByPageType` has nothing to do with the theme. I have a feeling we are not passing down all the required params for the endpoint which could be the reason it's breaking.

jairo-bc: I created a follow up ticket to investigate this (posted over slack in threads). There is a 500 error from graphql. 
Meanwhile, can we unblock with this temporary solution? 
cc @jmwiese @jkanive @zvuki

jmwiese: As long as we've captured the follow up work I think we can continue with this as a fix in the short term

jmwiese (approved):

PR #879, opened 2022-02-24
fix: windows circle ci

jairo-bc:
#### What?

Fix CircleCI build for windows, as we had _latest_ version in the setup and VS 2022 is in Preview.

https://github.com/actions/virtual-environments/issues/3949

### DIFF
--- a/.github/workflows/pull_request_review.yml
+++ b/.github/workflows/pull_request_review.yml
@@ -14,7 +14,7 @@ jobs:
         strategy:
             matrix:
                 node: [10.x, 12.x]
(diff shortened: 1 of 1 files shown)


jmwiese (approved):
The [README ](https://github.com/actions/virtual-environments#available-environments) on the virtual-environments repo indicates that `windows-latest` should still be pointing to `windows-2019`, however there is an open issue with others experiencing this problem as well so it seems like they are mid-migration for this.

https://github.com/actions/virtual-environments/issues/4856

PR #882, opened 2022-03-03
feat: allow `.css`-files imports within SCSS files

andreyvolokitin:
#### What?

[It's not possible currently to import raw css files (i.e. from `node_modules`) within a theme](https://support.bigcommerce.com/s/question/0D54O00006yPDn2SAG/include-plain-css-from-nodemodules-in-theme), because `stencil-cli` [provides no such ability](https://github.com/bigcommerce/stencil-cli/blob/master/lib/css-assembler.js#L26).

The standard for importing `.css` files in SCSS is to [not use any extension](https://sass-lang.com/documentation/at-rules/import#importing-css). The proposed change is to check for existence of `.css` file after all other options are exhausted. 

#### Tickets / Documentation

-   [This problem was recognised before](https://support.bigcommerce.com/s/question/0D54O00006yPDn2SAG/include-plain-css-from-nodemodules-in-theme)

### DIFF
--- a/lib/css-assembler.js
+++ b/lib/css-assembler.js
@@ -23,17 +23,22 @@ function assemble(cssFiles, absolutePath, type, options, callback) {
 
     const parseImports = (cssFileAlias, parseImportsCallback) => {
         const normalizedCssFileAlias = cssFileAlias.replace(/(\.scss)$/g, '');
-        const cssFilePath = path.join(absolutePath, normalizedCssFileAlias + ext);
+        const typedCssFilePath = path.join(absolutePath, normalizedCssFileAlias + ext);
+        const rawCssFilePath = path.join(absolutePath, normalizedCssFileAlias + '.css');
         const cssFileAliasDir = path.parse(normalizedCssFileAlias).dir;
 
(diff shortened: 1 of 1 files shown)


jairo-bc:
@andreyvolokitin Do you mind provide some proof (screenshots) that is working for css files and doesn't break previous logic (on cornerstone for example)?

andreyvolokitin:
Do you mean the theme being able to be built? Will do (I actually used this in exiting cornerstone theme), though the only difference with this PR is that the existing .css file will be included to the final css bundle, not sure how it can break previous logic?

jairo-bc:
@andreyvolokitin yes, just to double check if there's no regression coming up

andreyvolokitin:
Cornerstone gets successfully built and starts in dev mode, visually it doesn't look like there're any regressions: https://imgur.com/a/SeUbk9K

Also I think it's not possible to get CSS regressions from this PR, because all previous SCSS is handled as usual, and if absent it now brings as a fallback any `.css` with the same name. Any visual changes are only possible if deliberately introduced inside these CSS files

jairo-bc (approved):

PR #884, opened 2022-03-09
feat: strf-9440 Stencil Bundle: fail on scss failure compilation

jairo-bc:
#### What?

Scss compilation validation task during theme bundling by running `stencil bundle` command 

#### Tickets / Documentation

-   [STRF-9440](https://jira.bigcommerce.com/browse/STRF-9440)

Screenshot 2022-03-09 at 20 19 02

Screenshot 2022-03-09 at 20 47 32

### DIFF
--- a/lib/bundle-validator.js
+++ b/lib/bundle-validator.js
@@ -14,6 +14,7 @@ const ValidatorSchemaTranslations = require('./validator/schema-translations');
 const privateThemeConfigValidationSchema = require('./schemas/privateThemeConfig.json');
 const themeConfigValidationSchema = require('./schemas/themeConfig.json');
 const themeValidationSchema = require('./schemas/themeSchema.json');
+const cssCompiler = require('./css/compile');
 
 const VALID_IMAGE_TYPES = ['.jpg', '.jpeg', '.png', '.gif'];
 const WIDTH_COMPOSED = 600;
@@ -47,6 +48,7 @@ class BundleValidator {
             this._validateThemeSchema.bind(this),
             this._validateSchemaTranslations.bind(this),
             this._validateTemplatesFrontmatter.bind(this),
+            this._validateCssFiles.bind(this),
         ];
 
         if (!this.isPrivate) {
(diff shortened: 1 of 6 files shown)


server/plugins/theme-assets/theme-assets.module.js
```diff
@@ -62,35 +57,10 @@ internals.cssHandler = async (request, h) => {
...
-            browsers: configuration.autoprefixer_browsers,
-        },
-    };
-    const stencilStyles = new StencilStyles(console);
-    stencilStyles.activateEngine(SASS_ENGINE_NAME);
+    const themeAssetsPath = internals.getThemeAssetsPath();
 
     try {
-        const css = await stencilStyles.compileCss('scss', params);
+        const css = cssCompiler.compile(configuration, themeAssetsPath, fileName);
```
jmwiese, line 63: Do we need to `await cssCompiler.compile` here?

jairo-bc: yes!

jmwiese (approved):

lib/bundle-validator.js
```diff
@@ -402,6 +404,42 @@ class BundleValidator {
...
+        try {
+            const files = await fs.promises.readdir(stylesPath);
+            cssFiles = files.filter((file) => {
+                return file.slice(-(compiler.length + 1)) === `.${compiler}`;
+            });
+        } catch (e) {
+            throw new Error(`Error: ${e.message}, while reading files from "${stylesPath}".`.red);
+        }
+
+        for await (const file of cssFiles) {
```
jmwiese, line 423: Does the use of `for await` here mean that `cssFiles` is an  async interable object? It seems like we're doing `await` inside of the `try` so dose that mean the filter/slice operation results in `async`?

jairo-bc: I think it was a leftover from previous solution. But you are right, that `for await` should be used for async iterable objects. On other hand it also supports `sync` iterables, I think it's cleaner to separate those concerns

jairo-bc: I think the problem was in the eslint as well: https://eslint.org/docs/rules/no-await-in-loop
But those operation steps should go in sync way, so I disabled that rule for the next line

mattcoy-arcticleaf:
@jairo-bc , sorry, this appears to now be fixed in 4.1!

PR #886, opened 2022-03-10
fix: package-lock.json and ci workflow

jairo-bc:
#### What?

1) package-lock.json wasn't updated in the feature branch
2) run `npm ci` instead of `npm i` on opened PR


.github/workflows/pull_request_review.yml
```diff
@@ -39,7 +39,7 @@ jobs:
                       ${{ runner.os }}-node-
 
             - name: Install Dependencies
-              run: npm i
+              run: npm ci
```
jairo-bc, line 42: Don't update package-lock.json while running build on opened PR

PR #895, opened 2022-03-23
feat(storefront): bctheme-1063 modify stencil to start locally with components ui library

bc-alexsaiannyi:
#### What?

With this PR we would like to add opportunity to use templates or styles via external BigCommerce Lib( it will be created later). For this purpose we are going to implement a rule that indicates it's been taken from the external package and not from a directory within a Theme. 
For instance:
`{{> external/bc-ui-lib/templates/carousel-content }}`   **external/** is a flag that template comes from some external lib.
while 
`{{> components/common/carousel-content}}`. is a default way to use html component within a Theme.
#### Tickets / Documentation

-   [BCTHEME-1063](https://jira.bigcommerce.com/browse/BCTHEME-1063)

TBD

### DIFF
--- a/lib/template-assembler.js
+++ b/lib/template-assembler.js
@@ -6,6 +6,44 @@ const upath = require('upath');
 
 const partialRegex = /\{\{>\s*([_|\-|a-zA-Z0-9/]+)[^{]*?}}/g;
 const dynamicComponentRegex = /\{\{\s*?dynamicComponent\s*(?:'|")([_|\-|a-zA-Z0-9/]+)(?:'|").*?}}/g;
+const includeRegex = /{{2}>\s*([_|\-|a-zA-Z0-9/]+)[^{]*?}{2}/g;
+const packageMarker = 'external/';
+
+const isExternalTemplate = (templateName) => {
+    return templateName.startsWith(packageMarker);
+};
+const defineBaseRoute = (route) => route.split('/').slice(0, 3).join('/');
+const applyExternalPath = (templateName, templates) => {
+    return templates.map((t) => `${defineBaseRoute(templateName)}/${t}`);
+};
+const replacePartialNames = (content, partialRoute) => {
+    return content.replace(
+        includeRegex,
+        (__, partialName) => `{{> ${defineBaseRoute(partialRoute)}/${partialName} }}`,
+    );
+};
+
+/**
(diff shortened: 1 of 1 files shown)


BC-krasnoshapka (approved):

lib/template-assembler.js
```diff
@@ -7,6 +7,29 @@ const upath = require('upath');
...
 
+/**
+ * Takes a templates folder and template name. It returns simple path for custom template if it's available
+ *  or use default folder instead
+ *
+ * @param templatesFolder
+ * @param templateName
+ */
+function getCustomPath(templatesFolder, templateName) {
+    const packageFlag = 'external/';
```
jairo-bc, line 18: I believe it's not a flag. A `flag` is usually considerer a boolean variable.

lib/template-assembler.js
```diff
@@ -7,6 +7,29 @@ const upath = require('upath');
 const partialRegex = /\{\{>\s*([_|\-|a-zA-Z0-9/]+)[^{]*?}}/g;
 const dynamicComponentRegex = /\{\{\s*?dynamicComponent\s*(?:'|")([_|\-|a-zA-Z0-9/]+)(?:'|").*?}}/g;
 
+/**
+ * Takes a templates folder and template name. It returns simple path for custom template if it's available
+ *  or use default folder instead
+ *
+ * @param templatesFolder
```
jairo-bc, line 14: Do you mind adding a type here and below?

jairo-bc (changes requested):
Generally looks good. Left few suggestions. Also would you mind update Draft section in the changelog?

jairo-bc:
@bc-alexsaiannyi @BC-krasnoshapka I believe it's just a `stencil cli` part. Are you going to update `storefront-renderer` as well or it will be our work?

lib/template-assembler.js
bc-alexsaiannyi: it's changed to `packageMarker` since it indicates which source has been used for template. 
@jairo-bc  how do you find this name?

jairo-bc (approved):

jairo-bc:
is it @bc-alexsaiannyi fully tested and ready to be merged?

yurytut1993 (approved):

PR #904, opened 2022-04-07
feat(storefront): bctheme-1064 Modify Stencil bundle for using Components UI Library

yurytut1993:
#### What?
This PR provides functionality of adding templates from external lib to the theme bundle. The main idea is that script can understand is used template consist in the theme or inside external node_modules library.

IMPORTANT
It should be merged only after [bctheme-1063](https://github.com/bigcommerce/stencil-cli/pull/895)

#### Tickets / Documentation

[bctheme-1064](https://jira.bigcommerce.com/browse/BCTHEME-1064)

TBD


lib/stencil-bundle.js
```diff
@@ -49,6 +49,7 @@ class Bundle {
         this.options = options;
 
         this.templatesPath = path.join(themePath, 'templates');
+        this.externalTemplatesPath = path.join(themePath, 'node_modules/bc-ui-lib/templates');
```
jairo-bc, line 52: maybe it should come from config?
Have you thought about supporting multiple templates libraries?

lib/template-assembler.js
```diff
@@ -6,6 +6,44 @@ const upath = require('upath');
 
 const partialRegex = /\{\{>\s*([_|\-|a-zA-Z0-9/]+)[^{]*?}}/g;
 const dynamicComponentRegex = /\{\{\s*?dynamicComponent\s*(?:'|")([_|\-|a-zA-Z0-9/]+)(?:'|").*?}}/g;
+const includeRegex = /{{2}>\s*([_|\-|a-zA-Z0-9/]+)[^{]*?}{2}/g;
+const packageMarker = 'external/';
+
+const isExternalTemplate = (templateName) => {
```
jairo-bc, line 12: Could you please add here and below doc types?

lib/stencil-bundle.js
```diff
@@ -154,19 +155,35 @@ class Bundle {
...
 
-        recursiveReadDir(this.templatesPath, ['!*.html'], (readdirError, files) => {
-            if (readdirError) {
-                return callback(readdirError);
-            }
-
-            const partials = files.map((file) => {
+        // eslint-disable-next-line node/no-unsupported-features/es-builtins
+        Promise.allSettled([
+            recursiveReadDir(this.externalTemplatesPath, ['!*.html']),
```
jairo-bc, line 160: Will it work if the theme doesn't have external components library?

lib/template-assembler.js
```diff
@@ -6,6 +6,44 @@ const upath = require('upath');
 
 const partialRegex = /\{\{>\s*([_|\-|a-zA-Z0-9/]+)[^{]*?}}/g;
 const dynamicComponentRegex = /\{\{\s*?dynamicComponent\s*(?:'|")([_|\-|a-zA-Z0-9/]+)(?:'|").*?}}/g;
+const includeRegex = /{{2}>\s*([_|\-|a-zA-Z0-9/]+)[^{]*?}{2}/g;
+const packageMarker = 'external/';
```
jairo-bc, line 10: Maybe it's better to consolidate this name with  `externalTemplatesPath` in `stencil-bundle`?

lib/template-assembler.js
jairo-bc, line 12: Could you please add here and below doc types?

jairo-bc (changes requested):

lib/stencil-bundle.js
```diff
@@ -151,51 +152,98 @@ class Bundle {
         };
     }
 
-    assembleTemplatesTask(callback) {
+    async assembleTemplatesTask(callback) {
         console.log('Template Parsing Started...');
+        const internalTemplatesList = await recursiveReadDir(this.templatesPath, ['!*.html']);
+        const searchExternalLib = (content) => {
```
jairo-bc, line 158: Do you mind moving searchExternalLib and checkTemplate to class functions (and maybe renaming to make more sense in class domain). 
Maybe you can think also about extracting this functionality into separate file, while you are here?

lib/stencil-bundle.js
```diff
@@ -151,51 +152,98 @@ class Bundle {
...
-                    const ret = {};
+            return res;
+        };
 
-                    if (err) {
-                        return callback(err);
-                    }
+        let externalLibs;
+        try {
+            externalLibs = (await async.map(internalTemplatesList, checkTemplate)).flat();
```
jairo-bc, line 174: I would recommend not to use `async` library, where you can implement it in vanilla js

lib/stencil-bundle.js
```diff
@@ -151,51 +152,98 @@ class Bundle {
...
+            );
+
+            const ret = {};
+
+            allPartials.forEach((file, index) => {
+                ret[file] = results[index];
+            });
+
+            // eslint-disable-next-line no-unused-vars
+            const checkResult = await async.parallel([
```
jairo-bc, line 236: Here also. I know, that it was in the legacy code, but while you are working on this part, it would really helpful to reduce dependency from `async`

lib/stencil-bundle.js
```diff
@@ -151,51 +152,98 @@ class Bundle {
...
+                this._checkObjects.bind(this, results),
+                this._detectCycles.bind(this, results),
+            ]);
+
+            console.log(`${'ok'.green} -- Template Parsing Finished`);
+
+            return ret;
```
jairo-bc, line 243: did you try to use `return callback(null, ret)` here? What happened?

lib/stencil-bundle.js
```diff
@@ -317,56 +365,59 @@ class Bundle {
...
-
-                    if (failedTemplates.length) {
-                        return console.error(
-                            `Error: Your bundle failed as templates generated from the files below are greater than or equal to 1 megabyte in size:\n${failedTemplates.join(
-                                '\n',
-                            )}`,
-                        );
-                    }
+        async
+            .series(this.tasks)
```
jairo-bc, line 371: Is it because you removed the callback invocation?

jairo-bc (changes requested):

lib/stencil-bundle.js
yurytut1993: we received `task` object with field `templates` `undefined` which prevented us to create bundle

lib/template-assembler.js
yurytut1993: done

lib/stencil-bundle.js
yurytut1993: done

lib/stencil-bundle.js
yurytut1993: code was updated. no issue here

lib/stencil-bundle.js
yurytut1993: done

lib/stencil-bundle.js
yurytut1993: done

lib/stencil-bundle.js
yurytut1993: done

lib/stencil-bundle.js
```diff
@@ -152,26 +154,35 @@ class Bundle {
...
-            return externalTemplatesImports.map((_import) => _import.split('/')[1]);
-        };
-        const checkTemplate = async (templatePath) => {
-            const content = await fs.promises.readFile(templatePath, { encoding: 'utf-8' });
-            const res = searchExternalLib(content);
+        if (!externalTemplatesImports) return [];
 
-            return res;
-        };
+        return externalTemplatesImports.map((_import) => _import.split('/')[1]);
```
jairo-bc, line 172: Is there a reason using underscore in `_import`?

yurytut1993: do you have any good name suggestions?

lib/template-assembler.js
yurytut1993: We removed `externalTemplatesPath`
If your comment is still valid please can you add more details:)

lib/stencil-bundle.js
yurytut1993: np, updated

jairo-bc (approved):
LGTM

bc-alexsaiannyi (approved):

PR #910, opened 2022-04-11
feat: STRF-9706 Support node 14 and drop node 14

jairo-bc:
#### What?

Bump stencil styles, CI/CD to support node 14. Drop Node 10 

#### Tickets / Documentation

-   [STRF-9706](https://jira.bigcommerce.com/browse/STRF-9706)


jmwiese (approved):
I think the title should say "drop node 10" instead of 14

jairo-bc:
> I think the title should say "drop node 10" instead of 14

haha, yeah

PR #923, opened 2022-04-26
feat: STRF-9749 Update Stencil Cli to use node-sass latest by default

jairo-bc:
#### What?

Stencil bundle and stencil push will now validate scss file against latest node-sass version. If it fails, it's still possible to bypass the error by using` --use-old-node-sass-fork`

#### Tickets / Documentation

-   [STRF-9749](https://bigcommercecloud.atlassian.net/browse/STRF-9749)

Screenshot 2022-04-25 at 16 11 15


lib/bundle-validator.js
```diff
@@ -408,31 +411,100 @@ class BundleValidator {
...
                 /* eslint-disable-next-line no-await-in-loop */
-                await cssCompiler.compile(rawConfig, assetsPath, file);
+                await cssCompiler.compile(rawConfig, assetsPath, file, engine);
             } catch (e) {
                 const message = this.parseStencilStylesError(e);
+                if (!stencilCLISettings.isOldNodeSassForkEnabled()) {
+                    throw new Error(
+                        `${message}\n`.red +
+                            `\n` +
+                            `---------WARNING---------\n`.red +
```
jairo-bc, line 429: Are you okay with such wordings or do we need to adjust?

cc @bookernath

lib/bundle-validator.js
```diff
@@ -408,31 +411,100 @@ class BundleValidator {
...
+                            `We are currently in the process of deprecating node-sass fork https://github.com/bigcommerce-labs/node-sass\n` +
+                            `Your scss files were compiled using latest node-sass version https://github.com/sass/node-sass\n` +
+                            `This error might indicate that your scss file is not compatible with it.\n` +
+                            `There is still an option to compile scss file old fork by using --use-old-node-sass-fork.\n` +
+                            `But note, that this will lead to 500 error in production in near future.\n` +
```
bookernath, line 434:
```suggestion
                            `But note, that this will lead to 500 error in production in the near future.\n` +
```

lib/bundle-validator.js
bookernath: Looks good 👍  is there some link we can provide to help people diagnose errors? Some node-sass documentation perhaps?

jairo-bc: There are bunch of links, that you need to really dig to understand the context:
https://github.com/sass/node-sass/releases/tag/v3.5.1 - the release, that broke and changelogs.
https://github.com/bigcommerce/cornerstone/pull/999/files#diff-7b10e2a8164a3ef7ccb5aa035f393c9715b8945fcc59613d455408c505dd42a5L4 - known example of broken theme case (cornernstone)
https://github.com/sass/libsass/issues/1550 - code examples
https://github.com/sass/sass/issues/779 - code examples

@bookernath Maybe we can include all of them into out changelog and show the link for the changelog?

lib/cliCommon.js
```diff
@@ -88,11 +88,17 @@ function applyCommonOptions(program) {
     program
         .option('-h, --host [hostname]', 'specify the api host')
         .option('-nov, --no-verbose', 'supress verbose info logging', false)
+        .option(
+            '--use-old-node-sass-fork',
```
bc-max, line 92: looks like quite long option. What about adding some shortcut?
```suggestion
            '-oldn, --use-old-node-sass-fork',
```

CHANGELOG.md
```diff
@@ -1,9 +1,14 @@
 ### Draft
 
+-   feat: strf-9749 Update Stencil Cli to use node-sass latest by default ([x](https://github.com/bigcommerce/stencil-cli/pull/x))
+    New option:` --use-old-node-sass-fork` to bypass error warning (in case it exist) using latest node-sass version.
+    Soon, it node sass fork will be deprecated and it won't be possible to compile with it.
```
bc-max, line 5:
```suggestion
    Soon, node sass fork will be deprecated and it won't be possible to compile with it.
```

bc-max (approved):

lib/cliCommon.js
jairo-bc: @bc-max do you think `oldn` is enough descriptive? 
I was thinking about a shortcut, but didn't find a good one. Open for ideas :)

bc-max: > @bc-max do you think `oldn` is enough descriptive? I was thinking about a shortcut, but didn't find a good one. Open for ideas :)

Well, I think the description will be in documentation anyway. My understanding is that user is able to use both variants `-oldn` and `--use-old-node-sass-fork`. Finally, we have many shortcuts like `-h` or `-nov` 🤔

CHANGELOG.md
```diff
@@ -1,9 +1,14 @@
 ### Draft
 
+-   feat: strf-9749 Update Stencil Cli to use node-sass latest by default ([x](https://github.com/bigcommerce/stencil-cli/pull/x))
```
jmwiese, line 3: Should we include a proper PR number in this?

jmwiese (approved):

PR #953, opened 2022-07-04
fix(storefront): bctheme-1138 fix error on import components with @ symbol in the path

yurytut1993:
#### What?

This PR adds ability for theme developers to add components with `@` symbol in the path. 
For this component is required to wrap up its path with `' '` or `" "`
Any other components will work at the same way as it was previously

#### Tickets / Documentation

[BCTHEME-1138](https://bigcommercecloud.atlassian.net/browse/BCTHEME-1138)

importing

### DIFF
--- a/lib/template-assembler.js
+++ b/lib/template-assembler.js
@@ -4,9 +4,10 @@ const fs = require('graceful-fs');
 const path = require('path');
 const upath = require('upath');
 
-const partialRegex = /\{\{>\s*([_|\-|a-zA-Z0-9/]+)[^{]*?}}/g;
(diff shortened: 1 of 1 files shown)


lib/template-assembler.js
```diff
@@ -28,7 +28,12 @@ const replacePartialNames = (content, partialRoute) => {
         (__, partialName) => `{{> ${defineBaseRoute(partialRoute)}/${partialName} }}`,
     );
 };
-
+const trimPartial = (partial) => {
+    if (partial.startsWith(`'`)) {
```
jairo-bc, line 32: Do you want to support only single quotes? What about double quotes?

jairo-bc (approved):

bc-alexsaiannyi (approved):

PR #1033, opened 2022-11-23
feat(storefront): strf-9921 Hide logging for successful internal API calls in Stencil CLI:

bc-max:
#### What?

#### Tickets / Documentation

-   [Link 1](http://example.com)
-   ...

Example Image


lib/cliCommon.js
```diff
@@ -87,7 +87,7 @@ function checkNodeVersion() {
 function applyCommonOptions(program) {
     program
         .option('-h, --host [hostname]', 'specify the api host')
-        .option('-nov, --no-verbose', 'supress verbose info logging', false)
+        .option('-vb, --verbose', 'enable verbose info logging', false)
```
jairo-bc, line 90: Is it considered as a breaking change and should be documented? As we are following semver, we might want to add the note to the description (or title, we need to check in semantic release docs), so the tool automatically bumps the number to 6.x.x

cc @bookernath

jairo-bc (approved):

PR #1094, opened 2023-04-20
fix(storefront): MERC-9452 Return variation id when activating theme

bc-jcha:
#### What?

Currently Kibana and CP staff logs are logging variation ID when a theme is applied. Return a variation ID when activating a theme using Stencil CLI so that a developer can match the log with the corresponding variation ID.

#### Tickets / Documentation

https://bigcommercecloud.atlassian.net/browse/MERC-9452

Screen Shot 2023-04-21 at 1 26 16 PM

### DIFF
--- a/lib/stencil-push.utils.js
+++ b/lib/stencil-push.utils.js
@@ -515,7 +515,7 @@ utils.requestToApplyVariation = async (options) => {
 };
 
 utils.notifyUserOfCompletion = (options, callback) => {
-    callback(null, 'Stencil Push Finished');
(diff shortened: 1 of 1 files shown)


jairo-bc:
Hi @bc-jcha, 

Currently `chore` is not a valid commit type, but CI still passes. Do you mind changing it to fix, so we could publish can bump the version through CI?

jairo-bc (commented):
LGTM, just PR title needs to be changed

bc-jcha:
> LGTM, just PR title needs to be changed

Done, thank you! Approval would be appreciated @jairo-bc

kchu93 (approved):

PR #1114, opened 2023-06-16
Cypress error on load

isaacismaelx14:
#### What?
I noticed that when running Cypress for your E2E tests, it returns a 500 error. This occurs because for some reason, Cypress fails to include the `accept-language` header in the requests, resulting in an error during test execution.

To resolve this issue, I have made a modification by adding a validation step to check if `request.headers['accept-language']` is undefined before applying the `toLowerCase` method.

image

image

### DIFF
--- a/server/plugins/renderer/renderer.module.js
+++ b/server/plugins/renderer/renderer.module.js
@@ -373,7 +373,8 @@ function getAcceptLanguageHeader(request) {
     ) {
         return internals.options.storeSettingsLocale.default_shopper_language;
     }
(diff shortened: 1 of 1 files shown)


jairo-bc:
Hi @isaacismaelx14, thanks for contributing to the repo! The Accept Language Header is coming from the browser settings.  I think in order to have not empty, you should change that in the tests in Cypress or in the browser launcher (chrome, firefox, etc.)

jairo-bc:
@isaacismaelx14 Is it something you are working on?

PR #1183, opened 2024-04-08
feat: STRF-11741 Stencil init: auto install dependencies

jairo-bc:
#### What?

On Stencil init we are going to ask what's your package manager of flavour and install theme dependencies to eliminate additional command running

#### Tickets / Documentation

-   [STRF-11741](https://bigcommercecloud.atlassian.net/browse/STRF-11741)

Screenshot 2024-04-03 at 21 28 51
Screenshot 2024-04-03 at 21 29 02
Screenshot 2024-04-03 at 21 29 10

[STRF-11741]: https://bigcommercecloud.atlassian.net/browse/STRF-11741?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zvuki (approved):

lib/stencil-init.js
```diff
@@ -123,6 +132,16 @@ class StencilInit {
             });
         }
 
+        if (!cliOptions.packageManager) {
+            prompts.push({
+                type: 'list',
+                name: 'packageManager',
+                message: 'What is your favourite Package Manager?',
```
bc-max, line 139: The question itself sounds like a matter of taste. However, if I understood correctly we want to ask `What package manager do you want to use?`.  Sometimes answers on these 2 questions could be different. Just a thought.

bc-max (approved):

lib/stencil-init.js
jairo-bc: It is pretty much actually. Even if pnpm is the fastest, it's still a matter of taste. Actually, I took the question from Catalyst, so I think should be fine for now

chmelev:
This PR breaks automation because it does not allow supplying the packageManager value in the command line parameter, so it's just stuck waiting for the packageManager input.

salbahra:
This is a breaking change which broke our automation today. What was the purpose of this change?

PR #1222, opened 2024-08-28
Show channel IDs in channel prompt when starting dev server

loganwoolf:
#### What?

This simply shows the channel ID in the list of available channels when starting the server. Some themes may use the channel ID, so this smooths the experience of selecting the right channel.

#### Tickets / Documentation

None.

Screenshot 2024-08-28 at 5 42 26 PM


bc-max (approved):
💚 Works properly on MSF and non-MSF stores

LGTM! Good job 🎉

bc-max:
Need to revert this and recreate a PR due to incorrect commit message preventing from semantic release creation.

PR #1250, opened 2025-01-02
feat: Add port option to stencil start

onur-o7:
#### What?

Adds port option to stencil start. 

#### Why?

Our company manages multiple dev servers that each run on a unique port using config.stencil.port. It has become challenging to work efficiently on localhost as we need to track different ports. This way, when closing a tab or browser, we won't have to copy the URL or restart the dev server with stencil start -o again.

### DIFF
--- a/lib/stencil-start.js
+++ b/lib/stencil-start.js
@@ -55,7 +55,14 @@ class StencilStart {
         if (cliOptions.variation) {
             await this._themeConfigManager.setVariationByName(cliOptions.variation);
         }
+
         const initialStencilConfig = await this._stencilConfigManager.read();
+
+        // Override Port Number if Cli option Exsist
+        if (cliOptions.port) {
+            initialStencilConfig.port = cliOptions.port;
+        }
+
         // Use initial (before updates) port for BrowserSync
         const browserSyncPort = initialStencilConfig.port;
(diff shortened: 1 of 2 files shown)


lib/stencil-start.js
```diff
@@ -55,7 +55,14 @@ class StencilStart {
...
+
         const initialStencilConfig = await this._stencilConfigManager.read();
+
+        // Override Port Number if Cli option Exsist
+        if (cliOptions.port) {
+            initialStencilConfig.port = cliOptions.port;
+        }
+
         // Use initial (before updates) port for BrowserSync
         const browserSyncPort = initialStencilConfig.port;
```
jairo-bc, line 67: I think you can do instead
```suggestion
        const browserSyncPort = cliOptions.port || initialStencilConfig.port;
```

onur-o7: then we need do change 

```js
const updatedStencilConfig = this.updateStencilConfig(
            { initialStencilConfig, port: browserSyncPort},
            storeInfoFromAPI,
);

jairo-bc (approved):

jairo-bc:
@onur-o7 do you mind to rebase this one as well?

jairo-bc:
Released in https://github.com/bigcommerce/stencil-cli/releases/tag/8.9.0

Issue #1276, opened 2025-07-01
stencil-cli v8.8.0 doesn't bundle js, and doesn't create assets/dist


dragan0107:
I just tried this version of stencil and it fails to bundle the main theme js file, and it doesnt' even create assets/dist folder. I downgraded to 8.7.0 and it works just fine. I noticed you had a lot of version updates in the last couple of days, seems like something was messed up along the way.

Amish-Rokad:
@dragan0107 
"The issue was resolved by downgrading to an earlier version of Stencil CLI. Installing version 8.7.0 using the command below worked successfully:
**npm install -g @bigcommerce/stencil-cli@8.7.0**"

Amish-Rokad:
@dragan0107 
Hii 
Good Morning,

I am currently using Node.js v18.18.0
https://prnt.sc/gcOB-GoWP9KK

raviatelier:
@dragan0107 Thank you for the clarification.

Amish-Rokad:
Hello @dragan0107 @raviatelier,
If you have any work related to BigCommerce or Shopify, please feel free to let me know.
I’m a developer with experience in both platforms and would be happy to help!

dragan0107:
Thanks for the replies guys. I wanted to test the 8.8.2 version just now but according to the screenshots @sacr3dc0w posted, the issue remains, I guess we'll have to stick with the 8.7.0 for now. I'm also on the node 20.18.1. I'll have to do manual pushes for a while, since my project's pipeline installs latest stencil.

dragan0107:
Update: I just tested the latest version, 8.8.4, and they seem to finally have fixed the bundling problem 🙏

sacr3dc0w:
8.8.4 works, but only if I do `stencil bundle` first. Before, I could run `stencil push` and it worked.

dragan0107:
> 8.8.4 works, but only if I do `stencil bundle` first. Before, I could run `stencil push` and it worked.

@sacr3dc0w I just tested the `stencil push` with 8.8.4 and it worked, didn't have to `stencil bundle` first. I also see they published 8.8.5 in the meantime so you could try that one, and keep us updated.

dragan0107:
Pipeline just finished `stencil push` and the theme bundle JS file wasn't bundled, version 8.8.5. It was fixed with the 8.8.4 but they broke it again.

PR #1278, opened 2025-07-01
fix: STRF-000 Fix async config loading in theme bundling

AndrewBarber:
#### What?

Fixed async/await handling in theme bundling to ensure proper configuration loading.

#### Tickets / Documentation

This PR addresses issue #1276 where JavaScript files weren't being bundled and the `assets/dist` directory wasn't being created in stencil-cli v8.8.0. The root cause was improper async/await handling in the theme bundling process.

N/A


AndrewBarber:
@jairo-bc it looks like you've removed this feature - would this above be a fix for it instead? 
I think the issue lies in how we are awaiting for the `initConfig` function.

PR #1401, opened 2026-09-02
feat: require Node 24; BREAKING CHANGE: drop Node 20 and 22 support

parthshahp:
## Summary

Makes Node 24 the minimum supported version. This is a **breaking change**: Node 20 and 22 are no longer supported.

## Verification

On Node 24.13.1 / npm 11.8.0:
- `npm ci` passes
- `npm run lint` passes
- All 30 test suites / 222 tests pass
- `stencil bundle`, `stencil push`, `stencil start`, and `stencil init` all complete successfully

https://github.com/user-attachments/assets/c3abe209-a28a-4b1d-abf4-53d632b590d4


.prettierignore
```diff
@@ -4,4 +4,5 @@ node_modules
 .coverage/
 test/_mocks/
 
-CHANGELOG.md
\ No newline at end of file
+CHANGELOG.md
+parth/
```
jairo-bc: do we need it?

mise.toml
parthshahp, line 1: Deleted this because I don't think it's relevant for everyone, especially if they don't use mise.

jairo-bc (approved):

